PR #67, opened 2013-02-19
subtractDate incorrectly calculates date differences with different timezones.

kewisch:
`toUnixTime()` gets the unix time in UTC, but `subtractDate()` adds the utc offset again. Fix is quite easy, I'll upload a fix for this soon.


kewisch:
Hmm I just wanted the second commit to be in this PR, looks like I got the branches wrong.

kewisch:
Closing this PR to fix up the branches

PR #104, opened 2014-02-16
Add a clone method to ICAL.Period

kewisch:


kewisch:
@Taraman17, @gaye, @KevinGrandon can one of you take a quick look at this?

lib/ical/period.js
```diff
@@ -17,7 +17,8 @@
       this.start = aData.start;
     }
 
-    if (aData && ('end' in aData) && ('duration' in aData)) {
+    if (aData && 'end' in aData && 'duration' in aData &&
```
Taraman17: Why ist the check with "in" and .end/.duration needed?

kewisch: Yes, I think we can do without. I'll check if a strict warning is thrown and remove the "in" checks if not.

PR #106, opened 2014-04-04
Update to jCal final

kewisch:
jCal is now at the RFC Editor, so its time to update. Quite a few changes needed to be made, but most were fairly simple. All tests pass locally.

Note I also had to update the ical_previous file because the way the parser is being called has changed. This change is also backwards incompatible for consumers of the library, I'll update the wiki once this is merged.


kewisch:
@KevinGrandon @gaye @Taraman17 I'd appreciate if you could all take a look to make sure everything is as expected. You might have to change B2G a little too.

kewisch:
@gaye Can you take a look at this? I'd love to get this compatible to the jCal/jCard final.

kewisch:
I'm going to merge this now, its been too long :-)

PR #109, opened 2014-07-05
Fix hang on invalid newline inside quoted parameter

chotiwat:
A newline inside quoted parameter that is not the results of folding can cause ical.js parser to hang forever.

Example:

```
BEGIN:VCALENDAR
X-FOO;BAR="quoted
params";FOO=baz:realvalue
END:VCALENDAR
```

This pull request fixes the issue by throwing an error when a line of quoted parameter ends before the ending double quote is found.


kewisch:
Thanks for the report, patch and test!

chotiwat:
You're welcome. Thanks for the great library also!

PR #113, opened 2014-12-10
Make sandbox/validator.html work in Chrome

### LINKED ISSUE #112: Validator.html only works in Firefox
If you run the validator from Chrome you will get an error saying that toSource is not defined. I don't know if this is just intentional, if is you should probably mention that the validator is Firefox only in the readme.

Object.prototype.toSource
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/toSource

sunesimonsen:
Fixes #112 

The validator makes use of `toSource`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/toSource which is a non-standard method that only works in Firefox. The code that makes use of `toSource` seems to be dead, therefore I'm pretty sure that we can just remove the lines. I tested the changed code in Chrome and Firefox and works.

### DIFF
--- a/sandbox/validator.html
+++ b/sandbox/validator.html
@@ -35,7 +35,6 @@
         try {
             // First load the source representation
             sourceResult = ICAL.parse(data.value)
-            results.textContent = sourceResult.toSource();
 
             // Try to jsonify it
             jsonResult = JSON.stringify(sourceResult, null, " ");
@@ -57,9 +56,6 @@
             component = new ICAL.Component(sourceResult[1]);
 
             try {
-                // Try to undecorate it
-                undecorated.textContent = component.jCal.toSource();
-
                 // Try to jsonify it
                 jsonResult = JSON.stringify(component, null, " ");
                 undecorated.textContent = jsonResult;


kewisch:
The original intent was to first use toSource to get some representation of it, then try to JSONify it. If the jsonification fails because of undisplayable results or circular references, then it won't be filled in and just the toSource representation is left. Lets try this though and fix later if its a problem.

kewisch:
Can you also send a pull request for the gh-pages branch?

sunesimonsen:
Definitely.

sunesimonsen:
Hi again, actually it would be better to just point to 
https://cdn.rawgit.com/mozilla-comm/ical.js/master/sandbox/validator.html
from http://documentup.com/mozilla-comm/ical.js/ and the readme file and just delete the validator from gh-pages. I can fix the readme and delete the validator. but I can't update the wiki.

sunesimonsen:
In Chrome don't really like http://mozilla-comm.github.io/ical.js/, most of the links does not work, it seems Chrome is a bit more strict about some security areas than Firefox. Maybe it would be better just to redirect to http://documentup.com/mozilla-comm/ical.js/ using a redirection header: <meta http-equiv="refresh" content="0; url=http://documentup.com/mozilla-comm/ical.js" />

Sorry for laying all these suggestions on you.

kewisch:
Using rawgit sounds like a good idea to me, but I am not sure I understand your second comment. Does this mean using rawgit is not an option because it can't link to the ical.js build file?

No problem for the suggestions, I'm open to anything that will improve the experience :-)

sunesimonsen:
The gh-pages index.html is just an IFrame poiting to http://documentup.com/mozilla-comm/ical.js/ that make Chrome behave a bit weird, so I just suggested to redirect to the page from index.html instead.

Using rawgit means that we can just link to the validator.html from master and it will work.
https://cdn.rawgit.com/mozilla-comm/ical.js/master/sandbox/validator.html

This will remove the burden of keeping the gh-pages up to date.

PR #115, opened 2014-12-23
Fix #114

### LINKED ISSUE #114: Monthly recurring events appear on first day of month
Monthly recurring events without BYMONTHDAY occur on the first of the month, instead of the same day as the first occurrence.

See also https://bugzilla.mozilla.org/show_bug.cgi?id=1103187

This line looks suspect: https://github.com/mozilla-comm/ical.js/blob/6899a41b572de08d65ab544b51e0a93581bec197/lib/ical/recur_iterator.js#L761

darktrojan:
Patch by Decathlon, tests by me.

### DIFF
--- a/lib/ical/recur_iterator.js
+++ b/lib/ical/recur_iterator.js
@@ -612,10 +612,13 @@ ICAL.RecurIterator = (function() {
         }
 
       } else {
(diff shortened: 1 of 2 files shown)


kewisch:
Thanks for the patch and tests, great work!

Issue #117, opened 2014-12-30
Last day of a month in a yearly rule occurs on a wrong day


Decathlon1578:
This is bug 941425 on Mozilla comm-central.
With rules such as:
RRULE:FREQ=YEARLY;BYMONTHDAY=-1;BYMONTH=2
RRULE:FREQ=YEARLY;BYMONTHDAY=-1
the occurrences happen on wrong days and/or months.

kewisch:
Fixed by #118

PR #118, opened 2014-12-30
Fix recur yearly the last day of a month

Decathlon1578:
Fix for issue #117 with test about two cases.


kewisch:
Thanks for the PR and test!

PR #123, opened 2015-01-04
Remove fractional parts in adjust, not fromUnixTime

darktrojan:
Don't you hate it when you do something, then later think of a better way to do it? I do.


kewisch:
ICAL.Time::adjust is called quite often during timezone conversions and such, do you really think we should add the extra burden of truncating digits in ::adjust instead of doing it once in fromUnixTime?

darktrojan:
I suppose if someone decides to call adjust with dodgy arguments then it's their problem, fromUnixTime is much more likely to have a fractional part given that JS's date is in milliseconds. Let's leave it as it is.

PR #124, opened 2015-01-05
Fix recur yearly with Bymonth and more Byday

Decathlon1578:
It sorts the array of the days according to a Bymonthday order
(1,2,3, ...) instead of a Byday order (Su, Mo, ...).
Fixes Lightning's bug 958978 .

### DIFF
--- a/lib/ical/recur_iterator.js
+++ b/lib/ical/recur_iterator.js
@@ -943,8 +943,6 @@ ICAL.RecurIterator = (function() {
             }
           } else {
             for (var daycodedkey in this.by_data.BYDAY) {
-              //TODO: This should return dates in order of occurrence
-              //      (1,2,3, etc...) instead of by weekday (su, mo, etc..)
               var coded_day = this.by_data.BYDAY[daycodedkey];
               var parts = this.ruleDayOfWeek(coded_day);
               var pos = parts[0];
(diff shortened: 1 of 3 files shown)


lib/ical/recur_iterator.js
```diff
@@ -974,6 +972,9 @@ ICAL.RecurIterator = (function() {
             }
           }
         }
+        // Return dates in order of occurrence (1,2,3, etc...)
+        // instead of by weekday (su, mo, etc..).
+        this.days.sort(function(a,b){return a - b;});
```
kewisch: this.days only contains numbers, right? Do we actually need the custom comparator then?

Decathlon1578: Without comparator doesn't it sort alphabetically?

kewisch: Ah ok, so this.days is an array of strings?

Decathlon1578: Actually it's an array of numbers, for this reason sort() needs a comparator otherwise (without comparator) it sorts alphabetically and ascending. Am I wrong?

kewisch: Ah of course! I am wrong. I was testing with numbers 1,2,3 and obviously that didn't expose any issues!

kewisch:
Last thing before I merge, can you call `make package` so that the browser build is recreated?

lib/ical/recur_iterator.js
darktrojan: That probably should have an inline comment.

Decathlon1578:
Give me a bit of time, I've realized that git is an ugly beast!

kewisch:
No problem, take your time! Git took me a long time to get used to and there are still some things I don't like :)

kewisch:
Sorry, totally forgot about this one!

PR #128, opened 2015-02-24
Add caches to cacheable functions

darktrojan:
This has been sitting on my machine for a while, so I might as well add it.

### DIFF
--- a/build/ical.js
+++ b/build/ical.js
@@ -3319,9 +3319,6 @@ ICAL.TimezoneService = (function() {
     this.fromData(data, zone);
   };
 
-  ICAL.Time._dowCache = {};
-  ICAL.Time._wnCache = {};
-
   ICAL.Time.prototype = {
 
     icalclass: "icaltime",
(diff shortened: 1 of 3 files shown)


darktrojan:
Oh, damn, that test is failing because .day is directly manipulated. Might have to use getters and setters.

kewisch:
If at all possible, I think it would be good to keep all sorts of caching separate from the usual code. I'm not sure thats an option here though. Do you think we should add a way to disable the cache and/or limit its entries and evict entries?

What order of performance gain did you achieve with this?

darktrojan:
There's some nice gains
toUnixTime is about 4-5x as fast
dayOfWeek is about 1.5x as fast
weekNumber is about 22x as fast (shame we hardly ever use it)

… but it seems like the "add duration" test slows down a little. Investigating.

darktrojan:
If I set _cachedUnixTime to null instead of deleting it, the speedup is more in the order of 15x. TIL `delete` is expensive!

kewisch:
Ok, I am going to merge this. As mentioned I'm not too happy that the cache code is alongside the other code, but let go with it for now.

kewisch:
Oh, one more thing. Do you think you could add some tests that make sure the cache is evicted correctly when changes are made?

darktrojan:
I'm a bit busy at the moment but I'll try to find some time for that.

PR #129, opened 2015-02-24
Recurring events without dtend

ehourigan:
We came across an event from a google generated ical feed that has a 'vevent' with a 'rrule' set but does not have a 'dtend' property.  When calling event.getOccurrenceDetails(itr) the function errors trying to retrieve a duration.  Based on the ruleset below we have submitted this pr (with test) for review.

open to suggestions

thanks

for reference:
RFC 5545
3.6.1. Event Component
For cases where a "VEVENT" calendar component
specifies a "DTSTART" property with a DATE value type but no
"DTEND" nor "DURATION" property, the event's duration is taken to
be one day. For cases where a "VEVENT" calendar component
specifies a "DTSTART" property with a DATE-TIME value type but no
"DTEND" property, the event ends on the same calendar date and
time of day specified by the "DTSTART" property.


kewisch:
First of all, thank you for the patch with test! I have some minor comments, please stay tuned:

build/ical.js
```diff
@@ -6515,7 +6515,14 @@ ICAL.Event = (function() {
     },
 
     get endDate() {
-      return this._firstProp('dtend');
+      var end_date = this._firstProp('dtend');
+      if(end_date === null) {
```
kewisch, line 6519: What the original code was also missing is to handle the case where there is a DTSTART and a DURATION but no DTEND. Aside from changing the endDate getter, handling that case would also require changing the setter for endDate. The getter for duration should be fine because changing endDate would account for that, we can optimize this for performance later if its an issue.

build/ical.js
```diff
@@ -6515,7 +6515,14 @@ ICAL.Event = (function() {
...
+        end_date = this.startDate.clone();
+        if(this.startDate.isDate) {
+          end_date.addDuration(ICAL.Duration.fromSeconds(86400));
```
kewisch, line 6522: Some minor style nits: Could you add spaces after `if` and use a camelCase variable name?
Also, maybe it would be faster to just add 1 day to the end date using the day property, instead of constructing a duration object just to add the duration?

samples/recur_no_dtend.ics
```diff
@@ -0,0 +1,55 @@
+BEGIN:VCALENDAR
+PRODID:-//Google Inc//Google Calendar 70.9054//EN
+VERSION:2.0
+CALSCALE:GREGORIAN
+METHOD:PUBLISH
+X-WR-CALNAME:http://nhlc.org/32%2C35%2C37%2C38%2C40%2C41%2C42%2C43%2C45%2C4
```
kewisch, line 6: Could you remove the data identifying any external sites, i.e. drop the `X-WR-CALNAME` and change the `UID` to something else?

kewisch:
Funny, when I was reviewing pull request #138 I was pretty sure I had made the same review comments before. It seems you made the exact same change with slightly differing style. I don't really mind which is merged as long as all review comments are handled. @sunesimonsen @ehourigan

kewisch:
Fixed by #138

PR #132, opened 2015-02-26
Support of YYYY-MM-DD format in date/datetime.fromICAL()

laurentj:
Some vcards (probably generated by other tools than ICalendar)
may content date in the format YYYY-MM-DD, and not YYYYMMDD.

### DIFF
--- a/build/ical.js
+++ b/build/ical.js
@@ -501,16 +501,19 @@ ICAL.design = (function() {
         },
 
         fromICAL: function(aValue) {
-          // from: 20120901
-          // to: 2012-09-01
-          var result = aValue.substr(0, 4) + '-' +
-                       aValue.substr(4, 2) + '-' +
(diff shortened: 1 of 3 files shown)


kewisch:
This is probably vcard 3.0. The downside to this patch is that it would also accept the date format when converting iCalendar, which won't be correct. I think it might make more sense to write a separate preprocessor that would convert vcard 3.0 into vcard 4.0, and then have the parser concentrate on that. That separate preprocessor is welcome to live in ICAL.js though. What do you think?

kewisch:
I don't know where I got that from, although I do remember reading a vcard3 property line somewhere. Anyway, vcard4 also has date-and-or-time, and since 4dde47bee73cefba6fa32ebfc6dfde429bd0c0a0 we support that. Please reopen this PR if its still needed.

PR #138, opened 2015-03-13
Correct handling of dtend and duration for vevents

sunesimonsen:
Corresponding to https://tools.ietf.org/html/rfc5545#page-54 it is
allowed to supply a duration instead of a dtend and leave both of them
out. If there is only a dtstart and it is a date, it should be treated
an all day event. If there is only a dtstart and it is a time it
indicates that point in time.

This pull request fixes all the cases.


build/ical.js
```diff
@@ -6526,6 +6536,11 @@ ICAL.Event = (function() {
     },
 
     get duration() {
+      var duration = this._firstProp('duration');
+      if (duration) {
+          return duration;
+      }
+
```
kewisch: Could you flip this around so it matches the flow that get endDate has, i.e.

``` javascript
if (!duration) {
  return this.endDate.subtractDate(this.startDate);
}

return duration;
```

build/ical.js
```diff
@@ -6518,7 +6518,17 @@ ICAL.Event = (function() {
...
     get endDate() {
-      return this._firstProp('dtend');
+      var endDate = this._firstProp('dtend');
+      if (!endDate) {
+          var duration = this._firstProp('duration');
+          endDate = this.startDate.clone();
+          if (duration) {
+              endDate.addDuration(duration);
+          } else if (endDate.isDate) {
+              endDate.addDuration(new ICAL.Duration({ days: 1 }));
```
kewisch: Instead of constructing a new `ICAL.Duration` just for the purpose of incrementing the day, could you just increment using the `day` property?

samples/duration_instead_of_dtend.ics
```diff
@@ -0,0 +1,39 @@
+BEGIN:VCALENDAR
+PRODID:-//Google Inc//Google Calendar 70.9054//EN
+VERSION:2.0
+CALSCALE:GREGORIAN
+X-WR-CALNAME:calmozilla1@gmail.com
```
kewisch: Is this one of your addresses? If not, please use something at example.com to avoid adding real addresses in the source.

samples/duration_instead_of_dtend.ics
```diff
@@ -0,0 +1,39 @@
...
+TZNAME:PST
+DTSTART:19701101T020000
+RRULE:FREQ=YEARLY;BYMONTH=11;BYDAY=1SU
+END:STANDARD
+END:VTIMEZONE
+BEGIN:VEVENT
+DTSTART;TZID=America/Los_Angeles:20120630T060000
+DURATION:P1D
+DTSTAMP:20120724T212411Z
+UID:dn4vrfmfn5p05roahsopg57h48@google.com
```
kewisch: Same here, I think its better to use example.com

samples/only_dtstart_date.ics
```diff
@@ -0,0 +1,38 @@
+BEGIN:VCALENDAR
+PRODID:-//Google Inc//Google Calendar 70.9054//EN
+VERSION:2.0
+CALSCALE:GREGORIAN
+X-WR-CALNAME:calmozilla1@gmail.com
```
kewisch: Same comments on domain here and in all other samples.

kewisch:
Sorry for the delay. r=me with above changes.

samples/duration_instead_of_dtend.ics
sunesimonsen: Sorry about that, I did just copy the content from samples/minimal.ics

sunesimonsen:
I'll fix the problems and get back to you later today.

sunesimonsen:
@kewisch I fixed the problems you highlighted and force pushed the branch.

sunesimonsen:
Thanks

PR #139, opened 2015-03-16
Byday first month

Decathlon1578:
Patch for Lightning's bug 958974 and bug 419490.

### DIFF
--- a/build/ical.js
+++ b/build/ical.js
@@ -4714,33 +4714,43 @@ ICAL.RecurIterator = (function() {
       }
 
       if (this.rule.freq == "MONTHLY" && this.has_by_data("BYDAY")) {
-
-        var coded_day = this.by_data.BYDAY[this.by_indices.BYDAY];
-        var parts = this.ruleDayOfWeek(coded_day);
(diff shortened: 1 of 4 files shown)


Decathlon1578:
Not clear to me what went wrong.
Does Travis CI make tests with monthly recurrence rules with BYDAY part and weekdays with position part greater than 6?
Should we treat differently a BYDAY=6xx in a monthly rule?

kewisch:
Travis runs the same thing that you can run locally, although its running them through node and not via the browser. This usually doesn't make much of a difference though. I've run the tests with this PR applied and got the same failure both with node and browser tests.

It seems to be that the "Malformed values in BYDAY part" test in line 75 of test_recur.js is not throwing an error, although the test suggests it should be.

Decathlon1578:
My fault, I didn't run test_recur.js.
Since the patch now considers valid a rule BYDAY=+/-5XX, I've changed the test in test_recur.js and set a rule BYDAY=-6TH, that definitely is malformed for a monthly rule.

PR #147, opened 2015-04-08
Add support to TODO component and attribute allday for Event

yuniers:
TODO is a component of RFC 5545 that can be supported by ical.js
All day attribute represent that those event start and end on same day.
Too I added action attribute for alarms support.


build/ical.js
```diff
@@ -808,6 +811,7 @@ ICAL.design = (function() {
       },
 
       "action": DEFAULT_TYPE_TEXT,
+      "allday": DEFAULT_TYPE_BOOLEAN,
```
kewisch: allday is not expressed through an extra property, but instead by using the date fields with VALUE=DATE.

build/ical.js
```diff
@@ -6668,6 +6681,416 @@ ICAL.Event = (function() {
   return Event;
 
 }());
+
+ICAL.Todo = (function() {
```
kewisch, line 6618: I assume a lot of this is copy/paste from ICAL.Event? If so, it would probably make sense to create a BaseItem class. I don't think the BaseItem class needs to be exported on the ICAL object unless its needed to keep it in a separate file.

build/ical.js
```diff
@@ -6668,6 +6681,416 @@ ICAL.Event = (function() {
...
+
+    get startDate() {
+      return this._firstProp('dtstart');
+    },
+
+    set startDate(value) {
+      this._setTime('dtstart', value);
+    },
+
+    get endDate() {
```
kewisch, line 6909: todo items don't have DTSTART/DTEND, this needs to be adapted to what the RFC says.

kewisch:
Note also that `build/ical.js` is generated from the files in `lib/`. What you need to do is modify the files in `lib` and then use `make package` to compile `build/ical.js`

yuniers:
I don't know about it, form now I will change each file and later make 
package
Thanks for clarification.

build/ical.js
yuniers, line 6909: I think that a todo or task for me yes have _DTSTART_ but _DTEND is 
changed by DUE_ (present).

build/ical.js
yuniers, line 6618: Yes, I don't know that first must add a todo file and then do make package.

kewisch:
If you need help on fixing the above comments please let me know :)

yuniers:
Hi Philipp, on this days I'm was very busy and been work on others 
features. At night, I will start work on TODO.
I have a dude, how I can do for generate an organizer with name? I 
explain to you:
When I parse an ICS, the attribute organizer only have a mailto item and 
never show CN (if is present).

To, when an ICS is generated, the organizer is created in this way:
ORGANIZER:mailto:email. when should be _ORGANIZER;CN=Test;mailto:email_.

For parse and generate a correct organizer, we can create other file 
called organizer with same logic of recurs.js?

kewisch:
If its not possible using the `ICAL.Event` interface, you can always do it using the underlying component:

```
var event = new ICAL.Event();
event.organizer = 'mailto:email';

var organizerProp = myEvent.component.getFirstProperty('organizer');
organizer.setParameter('cn', 'Test');

console.log(event.toString());
```

kewisch:
I'm going to close this PR for now as some work is still required. I'd be delighted if you could find some time to fix the comments though. Note also, future commits will require jsdoc.

yuniers:
I have done that you advice to me but I don't know why not pass travis build.
It is my [todo.js](https://github.com/yuniers/ical.js/blob/master/lib/ical/todo.js) file.

kewisch:
Can you double check you got all comments? Replying from mobile email right now, but afair the base class was missing, and the date property names were not quite correct for tasks.

PR #154, opened 2015-05-12
improved test coverage for recur iterator

stephandenker:


kewisch:
Patch has been pre-reviewed via email. Thanks for the patch!

PR #49, opened 2015-05-25
Create API doc site.

kewisch:
The code base has more and more docs, lets use YUIDOC + the tools from my marionettejs project to add an API doc site.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ node_modules
 tools/tzurl
 tools/libical
 zoneinfo
+api
 coverage/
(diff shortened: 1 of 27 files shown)


kewisch:
I spent some time updating our inline docs, then generated a html page. I haven't yet set up travis to auto-push to gh-pages, but here is a preview: http://mozilla-comm.github.io/ical.js/api/

kewisch:
How awesome, after two more slight fixes, it works now. API documentation will always be up to date now!

PR #171, opened 2015-06-01
Fix for..in loop with modified prototype

amrod-:
Code in for..in loops break when ical.js is used in combination with code that adds functions to the prototype of Arrays. A simple hasOwnProperty check in the loops solves that this.

### DIFF
--- a/build/ical.js
+++ b/build/ical.js
@@ -710,21 +710,22 @@ ICAL.design = (function() {
         var str = "";
         for (var k in data) {
           /* istanbul ignore else */
-          if (Object.prototype.hasOwnProperty.call(data, k)) {
-            var val = data[k];
-            if (k == "until") {
-              if (val.length > 10) {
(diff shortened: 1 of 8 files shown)


kewisch:
Interesting, which Libraries are causing this? Ideally, those libraries should be using `Object.defineProperty` to make those functions non-enumerable properties.

For short loops with just 2-3 lines its ok, but for long loops I'd prefer using `continue` to skip the loop if its not an own property. This also avoids the extra indentation level. Do you think you could change that?

amrod-:
I'm using emberjs, having them non-enumerable would indeed be the preferred option.

The code is changed to use <code>continue</code> in long loops, if more needs to be changed just let me know.

kewisch:
Looks good now, thanks for the patch! I hope ember fixes their prototypes in the future.

amrod-:
Thanks for merging.

Issue #91, opened 2013-06-24
YEARLY recurrence on Feb 29th should not show on Mar 1st in non-leapyears


Taraman17:
A recurrence on 29th Feb each year like this:
RRULE:FREQ=YEARLY
DTSTART;TZID=Europe/Berlin:20120229T120000
DTEND;TZID=Europe/Berlin:20120229T130000

shows up on Mar 1st in non-leapyears.

Since the ical-spec says, that BYMONTHDAY parts > daysInMonth should not show, this should also apply to this special case.

wsdookadr:
@Taraman17 , @kewisch 

Hi, I've read this bug report.

My understanding is that from this list below, 2012 and 2016 are leap years
and the 29th of February is a valid month day in those two years(only
in those 2 years):
- **"2012-02-29T12:00:00"**
- "2013-03-01T12:00:00"
- "2014-03-01T12:00:00"
- "2015-03-01T12:00:00"
- **"2016-02-29T12:00:00"**
- "2017-03-01T12:00:00"
- "2018-03-01T12:00:00"

I've written the following test. I'm interested to know whether this is a proper test
for the bug described in this issue:

```
testRRULE('FREQ=YEARLY;', {
  dtStart: '2012-02-29T12:00:00',
  max: 7,
  dates: [
    '2012-02-29T12:00:00',
    '2016-02-29T12:00:00'
  ]
});
```

Currently the test fails since the iterator produces dates for each year
between 2012 and 2018, whereas from what I gather, it should only generate
two days, one in 2012 and one in 2016. The 1st of March that @Taraman17
mentions does indeed come up.

Is the test file **test/recur_iterator_test.js** (in the "YEARLY" suite) the right spot
for this test ?

@Taraman17

> BYMONTHDAY parts > daysInMonth

I wasn't able to find any mentions of this in [RFC5545](http://tools.ietf.org/html/rfc5545) .
Can you expand a bit on how BYMONTHDAY is involved in this example ?
Should your example contain BYMONTHDAY as well ?

Thanks

kewisch:
You are specifying "max: 7", so its generating 7 instances. If you would only like to check two instances, remove the max property completely. You also don't have to specify dtstart if the first instance in the dates array should be the dtstart.

But aside from that, if the test is generating instances for 2013, 2014, 2015, 2017, etc then it definitely shows the bug. It's in the right place in the yearly suite, if you want you can send a PR that adds the test but commented out feel free.

[...]

PR #175, opened 2015-06-04
added test for #91

wsdookadr:
This pull request contains a test for #91 .Since it's a test for an existing bug, the travis build for this will fail.

### DIFF
--- a/test/recur_iterator_test.js
+++ b/test/recur_iterator_test.js
@@ -949,6 +949,22 @@ suite('recur_iterator', function() {
           '2013-03-02',
         ]
       });
+
+      /* 
+       * New test for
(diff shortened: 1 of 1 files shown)


kewisch:
Out of curiosity, how did you manage to create a pull request from "unknown repository" ? As mentioned in the other issue, I think the best way forward is to change the nits I mentioned there and merge this test commented out. This way we won't forget to take care of it when moving to the new recurrence iterator.

wsdookadr:
@kewisch I've deleted the fork in the meantime, that's why it says "unknown repository".
I've made a [new pull request here](https://github.com/mozilla-comm/ical.js/pull/176).

PR #180, opened 2015-06-21
Fix weekOneStarts() according to ISO 8601

th-we:
I noticed

```
ICAL.Time.weekOneStarts(2015, ICAL.Time.MONDAY).day;
```

used to return `5`, which is wrong.  According to ISO 8601, Thursday 1st 2015 is in week one because it's the first Thursday of the year.  As ISO weeks start with Mondays, week 1 starts with the preceding Monday, which is December 29th 2014.  The above line should therefore return `29`.

As ISO weeks start with Mondays and the documentation for `weekOneStarts()` states "Returns the date on which ISO week number 1 starts", I changed the default for `aWeekDay` to Monday.

I see one remaining question:  What does it mean if `aWeekStart` is not Monday?  The ISO definition does not work any longer.  My fix returns the first day of ISO week 1 that is the supplied `aWeekDay`.  I'm not sure whether that's intended.

### DIFF
--- a/build/ical.js
+++ b/build/ical.js
@@ -5689,6 +5689,8 @@ ICAL.TimezoneService = (function() {
    * @return {ICAL.Time}            The date on which week number 1 starts
    */
   ICAL.Time.weekOneStarts = function weekOneStarts(aYear, aWeekStart) {
+    // ISO 8601 defines: Week 1 contains the first Thursday of the year.
+    // Equivalent: Week one contains Janury 4th
     var t = ICAL.Time.fromData({
       year: aYear,
       month: 1,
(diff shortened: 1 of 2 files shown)


kewisch:
Yes, I am aware of this issue. I've fixed it in https://github.com/kewisch/ical.js/commit/eecab591e2972c74b0d093930ae2c8d53ccd16be but I haven't ported this to master yet.

kewisch:
The first ISO week is always the week that contains the first Thursday of the year. This requires some shifting, but the way I interpreted that for other week starts it still works out. For example, if the week starts on wednesday, I still locate the first thursday of the year, then the week is from the wednesday before that (unless of course its the 1st).

kewisch:
I've cp'd the changeset in https://github.com/mozilla-comm/ical.js/pull/182, if this doesn't work for you please let me know there.

Issue #165, opened 2015-05-24
Assume the same default week start in all time functions


kewisch:
Some of the functions assume `ICAL.Time.SUNDAY` if no week start is passed, while the ISO 8601 default week start is Monday. I found at least one spot where we assume Monday. We should make sure to assume the same defaults in every function.

This will be a backwards incompatible change.

PR #182, opened 2015-06-21
Fix and thoroughly test ISO week number calculation

kewisch:


th-we:
This definitely works for ISO 8601 calculation - but only if Monday is supplied for `aWeekStart`.  I was fooled by the claim that `weekNumber()` "Calculates the ISO 8601 week number", which by default it still doesn't. I suggest to either let `aWeekStart` default to Monday for the functions claiming to be ISO conforming (i.e. `weekNumber()` and `weekOneStarts()`) if no argument is supplied, or change the documentation and give instructions how to use the functions to get ISO conforming results.

But while we're at it, I suggest to generalize the calculation. There are different definitions of what's the first week of the year ([Wikipedia](https://en.wikipedia.org/wiki/Week#cite_note-31) links to [this great resource](http://www.pjh2.de/datetime/weeknumber/wnd.php) in German). I'd suggest that one should not only be able to supply the week start, but also what date the first week needs to contain. The linked page suggests Jan. 1st, 4th and 7th are choices used officially somewhere in the world.

What do you think? I'd work that out.

th-we:
I thought about it and came up with something ([see comparison](https://github.com/mozilla-comm/ical.js/compare/master...th-we:new-weeknumber-logic)). I added some tests and modified the existing ones according to the logic I applied.

The problem is that this obviously breaks backwards compatibility with the "Thursday logic". But as calculation previously was buggy anyway, this might be less of an issue.

My question is whether the Thursday logic is in fact used in real life - outside ISO 8601, of course. It seems to me that all the systems in use (including ISO 8601) can be described by a certain day of January that needs to be in week 1.  I therefore introduced an argument `aJanNthInWeekOne`, which can also be set to 0 and will then use the system that allows incomplete weeks at the beginning and end of a year (i.e. Jan 1st is always in week 1, Dec 31st is always in last week).

kewisch:
Regarding passing `aWeekStart`, I had opened #165 to fix the default week start. We have a variable `ICAL.Time.DEFAULT_WEEK_START` which I think should be the default value for all functions that make `aWeekStart` optional, and the value for that variable should be Monday. If this is reflected correctly in the docs for `weekNumber` even better.

I like the idea of being able to support multiple week numbering systems. I think it would be great to pinpoint which systems use which dates, and provide predefined values that can be passed in, e.g.

```
tm.weekNumber(ICAL.Time.TUESDAY, ICAL.Time.WEEKSTART_ISO8601); // As specified, the default.
tm.weekNumber(ICAL.Time.TUESDAY, ICAL.Time.WEEKSTART_US); // Week 1 starts on January 1st, I believe.
```

The Thursday logic was easy to implement and seemed accurate. I don't think the week starts on days other than Saturday/Sunday/Monday in most parts of the world, but I'd like to have a system that works with any weekday, in theory. 

Let me know if I missed answering a question, need to run now!

th-we:
I'm happy about the positive response. I'm O.K. with either Sunday or Monday as day one, provided that it's consistent and clearly documented.

Yes, it would be good to have predefined settings - the question is, which ones are there and how to name them. I think it's sensible to start with ISO, US with complete and US with incomplete weeks (not sure whether we should call it US as it's also used in other countries around the world). If someone can confirm that another system is actually in use, this can easily be added.

Another question is how to organize those systems. I'd suggest introducing a new class `ICAL.WeekNumberSystem` and move the logic of `ICAL.Time.prototype.weekNumber` there.  I think to be fully flexible, it would be neat to have two ways of instantiating this class:
1. either supplying the properties the current logic requires (first day of week, day of January that needs to be in week one, flag signalling weeks can be incomplete) or
2. with a custom function that calculates week numbers. This would e.g. allow defining a function that keeps the Thursday logic for non-Monday based systems or entirely different logic like "week one always is Jan 1<sup>st</sup> to Jan 7<sup>th</sup>, week two is Jan 8<sup>th</sup> to Jan 14<sup>th</sup>" etc. or whatever peoples' requirements might be.

sunesimonsen:
I would really like this functionality as well.

kewisch:
I'm not yet convinced that we need a new class for this, I consider the week number stuff rather  a configuration, that can be specified as an object for the constant. If you can show me what per-system logic is required, I might reconsider :)

Looking at https://en.wikipedia.org/wiki/Week#Week_numbering there seem to be at least three different week numbering systems, I would start with these systems first. I agree "US" is the wrong name then, although it doesn't look like there is an official name as for the ISO standard.

Each week numbering system has its own default week start, so maybe instead of having the parameters `aWeekStart, aJanNthInWeekOne` we can flip them around and default to the week start specified in the constant if `aWeekStart` is not explicitly specified (e.g. when called from the recurrence iterator which takes it from RRULE's WKST). This would be a backwards incompatible change of course.

I could imagine this:

``` javascript
ICAL.Time.WEEKSTART_ISO_8601 = {
  nthJanuaryWeekOne: 4
  defaultWeekStart: ICAL.Time.MONDAY
};
ICAL.Time.WEEKSTART_NORTHAMERICA = {
  nthJanuaryWeekOne: 1
  defaultWeekStart: ICAL.Time.SUNDAY
};
```

PR #183, opened 2015-06-23
Vcard parser update

amrod-:
Fixed vcard 4.0 parser, added parsing for vcard 3.0


build/ical.js
```diff
@@ -340,7 +340,42 @@ ICAL.helpers = {
 ICAL.design = (function() {
   'use strict';
 
-  var ICAL_NEWLINE = /\\\\|\\;|\\,|\\[Nn]/g;
+  var FROM_ICAL_NEWLINE = /\\\\|\\;|\\,|\\[Nn]/g;
+  var TO_ICAL_NEWLINE = /\\|;|,|\n/g;
+  var FROM_VCARD_NEWLINE = /\\\\|\\,|\\[Nn]/g;
+  var TO_VCARD_NEWLINE = /\\|,|\n/g;
+
+  var textType = function(fromNewline, toNewline, type) {
```
kewisch: Since this function creates a text type, I think you should name it `createTextType`. Also, no need for variable assignment, you can just use `function createTextType(fromNewline, toNewline, type)`

build/ical.js
```diff
@@ -886,7 +894,9 @@ ICAL.design = (function() {
 
   // When adding a value here, be sure to add it to the parameter types!
   var vcardValues = ICAL.helpers.extend(commonValues, {
-
+    component: textType(FROM_ICAL_NEWLINE, TO_ICAL_NEWLINE, "text"),
```
kewisch: Sorry I missed this when I took an initial look, but can you tell me more about this new `component` type? I assume this is not a real type that will show up in the jCal, but instead `text` will show. Why do structured text values have different newline properties than normal text properties?

build/ical.js
```diff
@@ -1256,12 +1373,18 @@ ICAL.stringify = (function() {
...
     var result = 'BEGIN:' + name + LINE_ENDING;
-    designSet = designSet || design.getDesignSet(component[0]);
 
     var props = component[1];
     var propIdx = 0;
     var propLen = props.length;
 
+    var designSetName = component[0];
+    if (designSetName === 'vcard' && component[1].length > 0 &&
+            component[1][0][0] === "version" && component[1][0][3] === "3.0") {
```
kewisch: Can you add a quick comment here, noting that the rfc requires VERSION to be the first property?

build/ical.js
```diff
@@ -1256,12 +1373,18 @@ ICAL.stringify = (function() {
...
+      designSetName += "3";
```
kewisch: Same effect, but I'd prefer being explicit here and using `designSetName = "vcard3";`

kewisch:
When all comments are taken care of, can you rebase and squash the commits so its only one commit for all the changes, and one commit for the browser build?

build/ical.js
amrod-: In vCard 4.0 semicolon is not escaped in `text` values allowing lists which split on comma. However when there is a need for `lists` of `lists of text` semicolon is used as separator requiring it to be escaped, this value is called `component` in the rfc however the type value of it is still `text`. The name is a little bit misleading because it is a difference in escaping values than specific for newlines the variable can be renamed to something with ESCAPED in it.

amrod-:
Fixed the comments, rebased and squased.

build/ical.js
kewisch: So this is always the case when it is a `structuredValue`, right? Can you just make it depend on `structuredValue` being set, and then not escaping the character defined by `structuredValue`?

amrod-: Yes `component` is always in a `strcuturedValue`. That is also a way to solve this, not sure which one is the better solution.

kewisch:
Moving the `component` discussion here, so I don't always have to find the comments in the outdated diff. I think it would be cleaner to decide on escaping based on `structuredValue` (and maybe also `multiValue`, for that matter) instead of introducing a new "type" which is only a placeholder and not a real type. Can it be done without special casing for vcard3/vcard4/ical?

amrod-:
That can be done. The parse and stringify could tell the (to/from)ICAL function that they are in a structure. A simple `boolean` would suffice because the `structuredValue` is always a semicolon. The text type in vCard 4.0 can than decide to use an alternative regex (or modify the current one on the fly). Alternatively a function "structure(From/To)ICAL" can be added to the text type that parse/stringify uses when there is a structure, making it so no variable needs to be added to the (to/from)ICAL functions.

I don't think we need to take `multiValue` into an account here because no matter if it is used or not commas are always escaped in text.

kewisch:
I'm fine with a boolean if you prefer, but given `structuredValue` is configurable and not a boolean itself, I think we should pass it as a string and then have the to/fromICAL functions exclude that character.

amrod-:
Whoops, forgot to mention that I have updated the code so the <code>structuredValue</code> is used to exclude that character.

kewisch:
Heya, I totally forgot about this PR. I guess this one is good to merge and all previous comments are taken care of? I see code coverage has slightly decreased, can you look into that? One issue is that a comment says `istanbul ignore else` instead of `istanbul ignore if`, and I'm not sure about the UtcOffset lines. You don't have to fix lines that were not covered before of course

amrod-:
The previous comments were taken care of. I noticed that I had previously done the <code>istanbul ignore else</code> wrong in other places so I fixed those as well. A test for the UtcOffset is added in design_test.js.

PR #185, opened 2015-07-06
Make sure double imports work in node

kewisch:
```
var ical1 = require("ical.js");
var ical2 = require("ical.js");
// this causes ical2 to be {}
```


kewisch:
Note the current importing is really a mess. I'd love to make it easier, I believe I will have to make each file a module and switch to browserify for build/ical.js. I'll take care of this separately.

PR #192, opened 2015-09-16
Fixed bug where rrules saved to properties would be serialized incorrectly

sunesimonsen:
When setting a property value with a rrule it will undecorate the rrule, that
will translate the wkst from a number to a string. If the property later is
serialized to an ICAL string, the wkst field will be undefined.

### DIFF
--- a/build/ical.js
+++ b/build/ical.js
@@ -760,7 +760,9 @@ ICAL.design = (function() {
               val = icalValues.date.toICAL(val);
             }
           } else if (k == "wkst") {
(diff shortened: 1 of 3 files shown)


kewisch:
The failed checks are npm issues, I'll see if I can find the reason for them soon.

sunesimonsen:
Yes I could not figure out what they were about.

kewisch:
There we go, builds fixed. Thanks for the patch!

sunesimonsen:
Thanks a lot.

PR #196, opened 2015-11-01
Add support for multi value params which require DQUOTEs

make-my-day:
I removed the check in the parser for multiValueSeparateDQuote because this is within the dquoting section - in that case dquoting of the separate values is mandatory. The param getter/setter is currently used only at one place for timezones, but for consistency I added the a distinction for Array/String in/output.

### DIFF
--- a/build/ical.js
+++ b/build/ical.js
@@ -512,6 +512,12 @@ ICAL.design = (function() {
   var icalParams = {
     // Although the syntax is DQUOTE uri DQUOTE, I don't think we should
     // enfoce anything aside from it being a valid content line.
+    //
+    // At least some params require - if multi values are used - DQUOTEs
+    // for each of its values - e.g. delegated-from="uri1","uri2"
+    // To indicate this, I introduced the new k/v pair
+    // multiValueSeparateDQuote: true
+    //
     // "ALTREP": { ... },
 
     // CN just wants a param-value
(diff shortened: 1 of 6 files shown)


build/ical.js
```diff
@@ -1820,10 +1832,36 @@ ICAL.parse = (function() {
...
         pos = helpers.unescapedIndexOf(line, '"', valuePos);
+        if (multiValue && pos !== -1) {
+          var mvpos = pos;
+          var extendValue = true;
+          while (extendValue) {
+            var nextMDelim = helpers.unescapedIndexOf(line, multiValue, mvpos + 1);
+            var nextPDelim = helpers.unescapedIndexOf(line, PARAM_DELIMITER, mvpos + 1);
+            var nextVDelim = helpers.unescapedIndexOf(line, VALUE_DELIMITER, mvpos + 1);
+            var nextDQuote = helpers.unescapedIndexOf(line, '"', mvpos + 1);
+            if (nextDQuote > nextMDelim && (nextPDelim > nextDQuote || nextVDelim > nextDQuote)) {
```
kewisch, line 1857: Does this also work if some delimiter are not found? Also, is there a chance that you can use the last found position as a starting point for the next indexOf call to save a few ops?

kewisch:
Can you squash the commits when done with the comments? You can separate out the commit for the packaged build, but it should be two commits max.

kewisch:
Thanks for taking care!

PR #198, opened 2015-11-30
Fix #197, make UNTIL= use ICAL format, and make parsing fault-tolerant

### LINKED ISSUE #197: RRULE with UNTIL parsed incorrectly when date has - and :
I'm not sure which format is supposed to be used, but putting this

```
BEGIN:VCALENDAR
RRULE:UNTIL=20151231T060000Z
RRULE:UNTIL=2015-12-31T06:00:00Z
END:VCALENDAR
```

into the validator gives `2015-12-31T06:00:00Z` for the first and `2015--1-2-T1T:06::0` for the second. This is breaking recurrences for me in Lightning.

darktrojan:


kewisch (changes requested):
So, cleaning up some pull requests currently. I understand if you are no longer motivated to make changes, but do you think you could remove the code that makes the parser more lenient and just make sure the right date formats are used (if this is not already the case)?

I'd rather leave the migration part to the calling application and alternatively do it right by providing an option to make the parser more lenient.

kewisch (approved):

kewisch:
I've pushed this as bb10d62279ef1a1df97b9ef1ba7c549fed635649 so you don't have to rebase again.

PR #199, opened 2016-01-12
Bug 1208879 - JavaScript strict warnings about ReferenceError in ical.js

aleth:
These are the warnings that clutter the console on startup.

### DIFF
--- a/lib/ical/property.js
+++ b/lib/ical/property.js
@@ -211,7 +211,11 @@ ICAL.Property = (function() {
      * @return {Array|String}        Property value
      */
     getParameter: function(name) {
-      return this.jCal[PROP_INDEX][name];
+      if (name in this.jCal[PROP_INDEX]) {
+        return this.jCal[PROP_INDEX][name];
+      } else {
+        return undefined;
+      }
     },
 
     /**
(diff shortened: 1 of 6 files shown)


aleth:
@kewisch, would you mind running "grunt package" on this patch? I don't really want to install node etc just for a few undefined warning fixes if I can help it ;)

lib/ical/property.js
```diff
@@ -211,7 +211,11 @@ ICAL.Property = (function() {
      * @return {Array|String}        Property value
      */
     getParameter: function(name) {
-      return this.jCal[PROP_INDEX][name];
+      if (this.jCal[PROP_INDEX][name]) {
```
kewisch: I believe in strict mode this will still give you a warning, you'd need to do `typeof this.jCal[PROP_INDEX][name] == undefined` or `name in this.jCal[PROP_INDEX]`. Maybe you can also make it a ternary if to keep it a one-liner.

aleth: I tested it and it seemed to work (as we're not actually accessing the value of the undefined property), but |name in this.jCal[PROP_INDEX]| does seem more readable.

kewisch:
Let me know when you have the final patch ready, then I'll uglify/concat it for you. You'll need node eventually though ;-)

PR #200, opened 2016-01-12
Bug 1208879 - JavaScript strict warnings about ReferenceError in ical.js 

aleth:
@kewisch This one includes build/ical.js


aleth:
Isn't it nice how github simply removes the entire comment history when you squash? ;)

PR #202, opened 2016-01-30
Fix #201, Property parameters are misparsed when containing a second = sign

### LINKED ISSUE #201: Property parameters are misparsed when containing a second = sign
When I try to import the following .ics file: https://www.google.com/calendar/ical/asta.dhbw.de_c0g35t6hrh16kr4ankrqg2rdm4%40group.calendar.google.com/public/basic.ics parsing always breaks. One issue seems to be, that X-APPLE-STRUCTURED-LOCATION is not handled correctly, the other issue are '\n' linebreaks in the DESCRIPTION field. I created a jsfiddle with the latest ical data under https://jsfiddle.net/hupncr2y/ .

zoehneto:
Fixes #201. Also modifies the base64 test to prevent this bug in the future.


test/parser/base64.ics
```diff
@@ -1,13 +1,10 @@
 BEGIN:VCALENDAR
-ATTACH;FMTTYPE=text/plain;ENCODING=BASE64;VALUE=BINARY:TG9yZW
```
kewisch, line 2: Instead of changing the attachment value type to uri, I'd appreciate if you could keep it and then just add the parameter. Also, maybe you can clear out the real world data of that ics sample, both the real representation and the base64 encoded one, e.g.

```
ATTACH;FMTTYPE=text/plain;ENCODING=BASE64;VALUE=BINARY;X-BASE
 64-PARAM=UGFyYW1ldGVyCg==:WW91IHJlYWxseSBzcGVudCB0aGUgdGltZS
 B0byBiYXNlNjQgZGVjb2RlIHRoaXM/Cg==
```

zoehneto:
So now ci breaks on iojs because of some npm install issue :(

kewisch:
I've retriggered it, looks like a connectivity issue

kewisch:
Ok, checks passed. Can you rebase the changesets before I merge this?

zoehneto:
Rebase is done

PR #203, opened 2016-01-31
Include minified build and source maps in repository

kewisch:
It would be nice if the minified build and the source maps were included in the repository so a user installing ical.js using bower or npm already has the minified version (this seems to be in line with what most other projects do). It is also helpful considering the non minified build is around 260KB big.


kewisch:
Hmm yes, I considered this when I introduced the minifier. Ideally I'd like to be creating all of the build/ files only on releases, but this requires some hackery in creating a release branch for every release and only committing the files on the release branch.

Since build.js file is already being committed each time I might as well do the same for the minifier.

zoehneto:
That has the advantage that one could also tell bower to use the master branch or a certain commit without having to build the project, which would be necessary with a release branch.

kewisch:
Done, readme and contributing files have been changed.

PR #217, opened 2016-03-13
Automatically get exception event

strokyl:
I thought that it  will make sense, if when creating a Ical.Event the Ical.js library automatically populate the Event with exception by looking for them in the parent of the vevent component if it is present.
So here is my pull request.


lib/ical/event.js
```diff
@@ -49,6 +49,14 @@ ICAL.Event = (function() {
     if (options && options.exceptions) {
       options.exceptions.forEach(this.relateException, this);
     }
+
+    if (this.component.parent && !this.isRecurrenceException()) {
```
kewisch: I think we should expand this condition to make sure `options.exceptions` is not set. In this case the caller has specific intent w.r.t. the exceptions and may not want the parent exceptions to be automatically added or has already done so for legacy reasons.

Also, can you add some text to the header of the function that explains that exceptions will automatically be set in relation to the parent of the component, if the component is not a recurrence exception?

test/event_test.js
```diff
@@ -205,9 +205,7 @@ suite('ICAL.Event', function() {
     });
 
     test('with exceptions', function() {
-      subject = new ICAL.Event(primaryItem, {
-        exceptions: exceptions
-      });
+      subject = new ICAL.Event(primaryItem);
```
kewisch, line 208: I think it would be good to create a further test, making sure it works with both the automatic method and the manual method. Especially with the previously proposed changes.

kewisch (changes requested):
Sorry for the long delay :-( I hope you are still interested in making the following proposed changes:

lib/ical/event.js
strokyl: Done by transforming this if in a else if. I also updated the function documentation.

test/event_test.js
strokyl: I added a test, and I made the previous test's style better (IMO)

lib/ical/event.js
```diff
@@ -24,7 +24,7 @@ ICAL.Event = (function() {
    * @param {Boolean} options.strictExceptions
    *          When true, will verify exceptions are related by their UUID
    * @param {Array<ICAL.Component|ICAL.Event>} options.exceptions
-   *          Exceptions to this event, either as components or events
+   *          Exceptions to this event, either as components or events. If not specified exceptions will automatically be set in relation of component's parent
```
kewisch: Can you wrap this to 80 characters, indenting the next lines by two additional spaces?

kewisch (changes requested):
Almost there, just a minor nit! Thanks for the quick turnaround!

lib/ical/event.js
strokyl: done

kewisch (approved):

kewisch:
Thanks for the patience!

strokyl:
Great, when do you think you will release a new version of ICAL.js

kewisch:
I'll give it a little to see if the other PRs will have their changes made and then release a new version. In the meanwhile you can of course pin it to a certain version on master.

PR #218, opened 2016-03-14
Update RecurExpansion example

ghost:
Just a small correction


kewisch:
Thanks for taking care! Can you also run grunt package? If you don't want to set up grunt you can also change it in build/ical.js.

ghost:
Done :)

PR #229, opened 2016-04-13
ical.js@1.2.0 has grunt as a dependency

kewisch:
why? it's unnecessary and is breaking my build due to unmet peer dependencies. this seems to be caused by `grunt-node-inspector`.

also, i don't see v1.2.0 in git

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,9 +52,6 @@
     "sync-exec": "^0.6.0",
     "test-agent": "~0.28.2"
   },
-  "optionalDependencies": {
-    "grunt-node-inspector": "^0.4.1"
-  },
   "license": "MPL-2.0",
   "engine": {
     "node": ">=0.4"


kewisch:
Grunt is used for packaging, it is just a devDependency. I am in the process of updating to grunt 1.x but some of my devDependencies are not up to date yet. There are no normal dependencies.

1.2.0 is in git, see https://github.com/mozilla-comm/ical.js/tree/v1.2.0 and https://github.com/mozilla-comm/ical.js/releases/tag/v1.2.0

jonathanong:
screen shot 2016-04-13 at 1 52 30 pm

oh this is why. `optionalDependencies` is a "dependency" that is optional, not a "devDependency" that is optional.

jonathanong:
and it's confusing because v1.2.0 is tagged but not in master

kewisch:
Ah ok, good to know. I should remove grunt-node-inspector then, it looks like there is no optional devDependency. 

Looks like grunt release screwed up regarding the 1.2.0 release, I'll try to fix it.

kewisch:
This should do it. Anything else before I release 1.2.1?

PR #244, opened 2016-06-12
Support attachments

midzer:
Hi all.
With this small PR ical.js should handle calendar attachments and prepare data the same way as for ical event's member "attendees".

For more info on iCalendar attachments, see https://tools.ietf.org/html/rfc2445#section-4.1.3


kewisch:
Sorry for the delay! I understand that you would like to support attachments, but just returning the property won't be very helpful. I think we should wait until you or someone else has time to devise a good structure for a `ICAL.Attachment` high level object. In the meanwhile, you can do `event.component.getAllProperties('attach');`.

Nevertheless, thanks for creating a pull request for this!

PR #263, opened 2016-08-17
Bugfix days los on isDate change 

amrod-:
Example code:

```
var dur = ICAL.Duration.fromString("1D10H");
var start = ICAL.Time.fromString("2016-08-10T14:00:00");
start.addDuration(dur);
start.isDate=true;
var r1 = start.toString();

var dur = ICAL.Duration.fromString("1D10H");
var start = ICAL.Time.fromString("2016-08-10T14:00:00");
start.addDuration(dur);
start.isDate
start.isDate=true;
var r2 = start.toString();
```

r1 and r2 are different because isDate has an effect on the way time is normalized which should happen before isDate is set.


kewisch (changes requested):
Looks good to me, thanks for adding the test. Can you rebase this and generate a new packge? I'll merge it asap.

amrod-:
Done

kewisch (approved):

Issue #266, opened 2016-09-26
Fail to convert jCard to vCard


dbanetto:
Tested on master and version 1.2.2.

Was trying to convert a from a jCard to a vCard string and have been using the jCard provided
in the [RFC](https://tools.ietf.org/html/rfc7095#appendix-B.1.2)

**Expected behaviour**:

The Component will return a string containing a vCard.

**Actual behaviour**:
Throws an error in the browser console.

``` javascript
TypeError: comps is undefined ical.js:1389:9
    stringify.component() ical.js:1389
    Component.prototype.toString() ical.js:2642
    window.onload() _display:52
```

**Example**:

jsfiddle example: http://jsfiddle.net/gegxngqj/

kewisch:
This is indeed a bug in the library, I'm surprised this didn't come up earlier. Maybe it was a recent regression. The problem is that the component stringifier assumes that it is getting an object like: 

`["vcalendar", [ ...properties... ], [ ...subcomponents... ]`.

The vcard object is just:

`["vcard", [ ...properties... ] ]`

A simple fix would be to accept not passing subcomponents, by checking `components.length > 1` around the place where the error occurs. Some tests to ensure this doesn't happen again would be great too. The other direction should also be tested, making sure that getting the jCard data from a vCard string produces arrays so that there is no third member.

jCal requires the third element in the array to exist, even if it is an empty array. I can accept being lenient here, but ideally we'd indicate an error if the third element is missing for jCal.

Would you be willing to work on a fix for this?

dbanetto:
Thank you for the great explanation.
I'll have a go at fixing this and comment here if I run into any problems.

kewisch:
Thanks! Also feel free to catch me on IRC if you have questions. My nick is `Fallen` on irc.mozilla.org.

kewisch:
This was fixed in 2016, we just didn't close the issue. Thanks again!

PR #268, opened 2016-09-27
fix error when stringify'ing jCards without sub-components

dbanetto:
solves issue #266 

This includes some tests to ensure a jCal and the two different configurations of jCards will stringify correctly.

This does not solve the other problem brought up in the issue of jCard's made from parsing vCard string having an empty array for sub-components.

A potential problem with this fix is that a handcrafted/bugged jCal could be stringified without the required sub-components. A possible solution is checking the `designSet` if it is required.


build/ical.js
```diff
@@ -1384,12 +1384,16 @@ ICAL.stringify = (function() {
       result += stringify.property(props[propIdx], designSet) + LINE_ENDING;
     }
 
-    var comps = component[2];
```
kewisch, line 1387: I think it would be easier to read to just do `var comps = component[2] || [];`. r+ with that change.

kewisch (changes requested):

kewisch (approved):

kewisch:
I went ahead and pushed this as 955e4744a8c6563a5cba6fd81322179f70601ade so you don't have to make the changes. Thanks for the patch!

Issue #188, opened 2015-09-06
Feature: Helper function to clean up unreferenced timezones from icalendar object


untitaker:
When adding/removing DUE dates from Todo objects, timezone definitions can become obsolete. It'd be nice if there were helper functions available to manage those timezones (like the one described in the title).

kewisch:
Indeed, some code for timezone management would be nice. We have something like that in Lightning (where ical.js is also used) which is managed separately from ical.js. It would be great if we could hook into the code in a way that timezone management is not mandatory and does not complicate the code too much, while still providing the option to automatically add/remove vtimezones.

Note also, including VTIMEZONEs will no longer be mandatory in the future, see https://tools.ietf.org/html/draft-ietf-tzdist-caldav-timezone-ref-04

untitaker:
I'd like to see a function called `fixTimezones` that removes unused timezones and adds undefined ones completely automatically. I guess it should be somehow more configurable to adapt for that RFC, but my usecase has nothing to do with CalDAV.

wrexroad:
@kewisch should this issue be closed now that #270 has been merged?

untitaker:
Yes indeed. Thanks!

PR #270, opened 2016-10-13
Cleanup vtimezone

wrexroad:
Adds `ICAL.helpers.updateTimezones()` to fix issue #188.

The function takes an `ICAL.Component` as its only parameter and normalizes its `vtimezone`s in two steps. First, any vtimezone subcomponents that are not referenced by properties of any non-`vtimezone` subcomponents are removed. Then it adds any `vtimezone`s that are referenced but do not yet exist as a subcomponent (as long as the `vtimezone` has previously been registered with `ICAL.TimezoneService`).


kewisch:
Hi, sorry for the delay. I've been looking to implement something like this for a while but never got around to it. Thanks for taking care! I hope we can eventually move to ES6 Sets instead, but for now this looks great.

Would you mind rebasing this to build the latest ical.min.js and also squash commits so there is one for the whole feature and one for the browser build? I'll merge it asap.

kewisch (changes requested):
Marking this as requesting changes for github

kewisch (approved):

PR #278, opened 2016-12-03
mods based on vCard / vContacts work

neandr:


kewisch:
it seems you have made some changes to build/ical.js and some to the lib/ files. Can you make sure to change the lib/ files first then call grunt package? See docs for instructions on how to build.

build/ical.js
```diff
@@ -1073,7 +1073,7 @@ ICAL.design = (function() {
     "nickname": DEFAULT_TYPE_TEXT_MULTI,
     "note": DEFAULT_TYPE_TEXT,
     "org": { defaultType: "text", structuredValue: ";" },
-    "photo": DEFAULT_TYPE_URI,
+    "photo": { defaultType: "binary", allowedTypes: ["binary", "uri"] },             //XXXgW
```
kewisch, line 1076: If this is the right change, then please remove XXX. If it is temporary, please remove it.

build/ical.js
```diff
@@ -1134,7 +1134,16 @@ ICAL.design = (function() {
...
     n: { defaultType: "text", structuredValue: ";", multiValue: "," },
     nickname: DEFAULT_TYPE_TEXT_MULTI,
-    photo: { defaultType: "binary", allowedTypes: ["binary", "uri"] },
+ //   photo: { defaultType: "binary", allowedTypes: ["binary", "uri"] },
+ //   detect binary or http(s) link 
+    photo: { 
+      defaultType: "binary", 
+      allowedTypes: ["binary", "uri"],
+      detectType: function(string) {
+         return ((string.indexOf('https://') === 0) || (string.indexOf('http://') === 0) ? 'uri' : 'binary');
```
kewisch, line 1143: an uri is not necessarily http(s):// even though common. It could also be a data uri or something else. If type detection is really needed (i.e. lots of non-standard use) please look into a different check.

lib/ical/binary.js
```diff
@@ -155,7 +155,24 @@ ICAL.Binary = (function() {
...
       dec = tmp_arr.join('');
 
-      return dec;
+
+      // expand for codes > 128 
+      tmp_arr = []; cntrl = 0;
+      i =0, ac=0; 
+      do {
+        h1 = dec.charCodeAt(i++)
+        h2 = dec.charCodeAt(i)
```
kewisch, line 164: I think the i and i++ lines here may be a bit confusing. I'd suggest using a for loop and then indexing with `i` and `i+1`.

kewisch (changes requested):

neandr:
See comments, need advice for 'binary' check

kewisch:
I'm closing this PR for now as per the email I sent. Looking forward to a new version!

PR #302, opened 2017-04-07
RANGE:THISANDFUTURE is a parameter of Recurrence-ID, not a standalone property

georgehrke:
`RANGE` is only a [parameter](https://tools.ietf.org/html/rfc5545#section-3.2.13) of a [Recurrence ID](https://tools.ietf.org/html/rfc5545#section-3.8.4.4) and not a standalone property.


kewisch:
Looks right to me, thanks for the patch!

PR #310, opened 2017-05-05
Correct parsing of RRULE parts. Fixes #292

### LINKED ISSUE #292: Recursion error when parsing a VTIMEZONE in an English ICS, works fine in the same one in French
Hi @kewisch,

I hope you're doing well :)
We're still using ical.js in OpenPaas are it gives nice results. I do came on an issue that I couldn't explain, even after trying to understand relevant part of the code.

I have two ICS I receive from Outlook.com (https://outlook.live.com/owa/?path=/calendar/view/WorkWeek). One in english [INBOX-454-en.ics.txt](https://github.com/mozilla-comm/ical.js/files/804174/INBOX-454-en.ics.txt) and one in french [INBOX-454-fr.ics.txt](https://github.com/mozilla-comm/ical.js/files/804173/INBOX-454-fr.ics.txt). You'll notice that the only differences are in the TZID of the timezone and obviously in the DTSTART and DTEND properties that references the Timezone.

Parsing the french one and accessing individual properties works fine.
Parsing the english one triggers the below error

```
Error: Same occurrence found twice, protecting you from death by recursion
    at icalrecur_iterator_next [as next] (ical.js:5245)
    at icaltimezone._expandComponent (ical.js:3335)
    at icaltimezone._ensureCoverage (ical.js:3237)
    at icaltimezone.utcOffset (ical.js:3123)
    at icaltime.utc_offset [as utcOffset] (ical.js:4017)
    at icaltime.toUnixTime (ical.js:4185)
    at icaltime.toJSDate (ical.js:4063)
    at _calMoment (fc-moment.js:25)
    at CalendarShell.get end [as end] (calendar-shell.js:159)
    at eval (eval at zf (angular.js:12683), <anonymous>:8:4)
```

Do you have an idea what causes this?

Thanks a lot for your support,
Regards,

David

ddolcimascolo:
Hey @kewisch,

This fixes our issue on the 1.0.x branch of ical.js
I've no idea how to write tests for it, and to be honest I didn't even check if this broke anything :fearful: 

  
Details:

I've tracked down the issue to `icaltime_compare` which erroneously return `0` if one of the two compared Time instances have `NaN` as the unix time. How is that even possible, you'd say?
In our case, because when the recurrence iterator moves forward the years, it calls `increment_xxx` methods that uses the `interval` of the recurrence rule which is supposed to be an integer. However, it is a String in our case, I'm not exactly sure why...

Forcing a correct parsing in the `fromData` method of the timezone component fixed it, making sure interval is a integer.
  

Can you give me your insights and ideas how to properly test this?

Thanks,
David

### DIFF
--- a/build/ical.js
+++ b/build/ical.js
@@ -4728,6 +4728,8 @@ ICAL.TimezoneService = (function() {
           } else {
             this.parts[uckey] = [data[key]];
           }
+        } else if (uckey in optionDesign) {
+          optionDesign[uckey](data[key], this);
         } else {
           this[key] = data[key];
         }
--- a/lib/ical/recur.js
+++ b/lib/ical/recur.js
@@ -112,6 +112,8 @@
           } else {
             this.parts[uckey] = [data[key]];
           }
+        } else if (uckey in optionDesign) {
+          optionDesign[uckey](data[key], this);
         } else {
           this[key] = data[key];
         }


ddolcimascolo:
Wrong patch, sry.

ddolcimascolo:
Reopening with another tentative patch :)
Same comments apply

kewisch:
Hey David, thanks for upstreaming your patches! As for testing, do the instructions at https://github.com/mozilla-comm/ical.js#tests help?

PR #331, opened 2017-07-20
Add Validation for Bysetpos during initialisation

nkskalyan:
Attempted Fix for https://github.com/mozilla-comm/ical.js/issues/328

### DIFF
--- a/test/recur_iterator_test.js
+++ b/test/recur_iterator_test.js
@@ -1187,6 +1187,333 @@ suite('recur_iterator', function() {
         ]
       });
 
+      testRRULE('FREQ=DAILY;INTERVAL=2;COUNT=3', {
+        dtStart: '2013-01-01',
+        byCount: true,
+        dates: [
+          '2013-01-01',
+          '2013-01-03',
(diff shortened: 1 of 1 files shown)


kewisch:
Sorry I've been delaying this. Any changes to recurrence processing makes me feel uneasy. I'm glad you have tests and I *think* our test coverage is good, but I'm always a bit unsure. I'll see if I can pick this up again soon.

kewisch:
Wow, it has been like 8 years. 2 more and we could have had an anniversary :-} Looking at the code there are two concerns:
1) It appears that to validate you have another iterator running, which I believe could hurt performance. I think we'd need to find a different way to approach this.
2) My understanding of the RFC is that the DTSTART is always part of the occurrences, even if the RRULE does not match.

Given you've likely moved on from this I'm going to grab the tests that pass from your set and push them to the repo. My apologies for the experience.

kewisch (approved):

nkskalyan:
No worries and thank you! I used this for a job at the time and we ended forking it internally.

PR #337, opened 2017-08-19
Sort byparts before setting defaults in iterator

### LINKED ISSUE #336: Recurrence Missing instances if bymonthday is not sorted
```js
recur = ICAL.Recur.fromData({
 freq:"MONTHLY",
 bymonthday: [3, 2, 1],
 until: "2017-07-02"
}); // 1st, 2nd, 3rd of every month till 2017-07-02
start = ICAL.Time.fromDateString("2017-07-01") 
it = recur.iterator(start)
console.log('Expected 2017-07-01, Actual ' + it.next()); //null
console.log('Expected 2017-07-02, Actual ' + it.next()); // null
```
https://jsfiddle.net/jkcamnjx/3/

arunpandianp:
https://github.com/mozilla-comm/ical.js/blob/master/lib/ical/recur_iterator.js#L1372
setup_defaults is returning the first element of bypart array, if the input bypart array are not sorted then the iteration is not always correct. In this PR i'm sorting the bypart arrays before calling setup_defaults.  This also fixes #336


kewisch:
Thanks, I think this makes sense. Can you rebase on master? There were some conflicts in recur_iterator.js

build/ical.js
```diff
@@ -6974,12 +6974,26 @@ ICAL.RecurIterator = (function() {
...
       this.last.minute = this.setup_defaults("BYMINUTE", "MINUTELY", this.dtstart.minute);
+
+      this.sort_bypart_rules("BYHOUR", parts);
       this.last.hour = this.setup_defaults("BYHOUR", "HOURLY", this.dtstart.hour);
-      this.last.day = this.setup_defaults("BYMONTHDAY", "DAILY", this.dtstart.day);
+
+      this.sort_bypart_rules("BYMONTH", parts);
       this.last.month = this.setup_defaults("BYMONTH", "MONTHLY", this.dtstart.month);
 
+      //setting day after month, as sorting BYMONTHDAY will differ based on the month
```
kewisch, line 6990: If sorting bymonthday differs based on month, wouldn't we need to sort this differently every month while the rule is iteraed?

PR #344, opened 2017-12-08
make parsing time more flexible

jnrepo:
https://github.com/mozilla-comm/ical.js/issues/186

Branch with @fbacker 's solution


kewisch:
Fixed this with #412

PR #357, opened 2018-03-30
Property: introduce getFirstParameter method

sngo25:
When paramter is multivalue (categories, type, etc),
Property#getParameter will returns an array instead of string.
Hence, there will be the need of getting only string.


sngo25:
Hello @kewisch, how are you doing? I ran `grunt package` but CI still failling, can you have a look?

kewisch:
Yeah, the CI is a bit finicky. The minified build is usually different on each machine. I'm not too happy having the minified files in the tree, but not having them there makes it harder to reach the dist builds. I've run the packaging again, hope it works now.

mschroeder:
@heroandtn3: At the moment it is best to use yarn (not npm) to install the correct dependencies to build and package. Use "yarn --frozen-lockfile" to install all the node modules with the same versions as CI.

sngo25:
@kewisch @mschroeder it is passing now, thank you!

PR #361, opened 2018-08-07
Processor breaking when doing recurrence with the iterator

Aragorn450:
This is related to [https://github.com/mozilla-comm/ical.js/issues/360](https://github.com/mozilla-comm/ical.js/issues/360)

The issue that this solves is where an ical feed has the RRULE set to WEEKLY but also has the BYDAY that specifies which day (1st Friday for instance). In that case, the WEEKLY ignores which Friday is specified and instead treats it like every Friday.

Even worse, if the BYDAY has something like 1FR,2FR, the WEEKLY processing will again, ignore that it's the 1st and 2nd Friday and instead have each Friday output twice. Once for the 1FR and once for the 2FR.

By looking to see if there's a number and changing processing to MONTHLY (as it should have been specified in the ical feed), both issues are fixed.


Aragorn450:
Hmm, not sure why the travis-ci failed. It says it needs `grunt package` run but that WAS run before that second commit. And running it again gives no changed files...

kewisch:
Sorry I've been delaying this. Any changes to recurrence processing makes me feel uneasy. I'm glad you have tests and I *think* our test coverage is good, but I'm always a bit unsure. I'll see if I can pick this up again soon.

Can you add some info on the change of recurrence pattern? Preferably backed by 5545 in some way. I'll check the spec myself, but if you have some time and additional info I'd appreciate.

PR #389, opened 2019-04-05
[WIP] Adds support for vCard 2.1

### LINKED ISSUE #187: Support VCard 2.1
If I try to parse a vCard in 2.1 format, the parser does not accept it, since the "TYPE=" tags are missing. Is vCard 2.1 not supported?

caugner:
Fixes #187.

- [ ] [1] Support `QUOTED-PRINTABLE` inline encoding [cf. Section 2.1.2].
- [x] [2] Do not interpret "\n" or "\N" as a `CRLF`(but instead literally as "\\", followed by "n" or "N").
- [x] [3] Do not interpret `COMMA` as a multi-value separator [all values are single values].
-  [3] *Note*: Despite what vCard 3.0 states, a `SEMICOLON` must also be backslash escaped in vCard 2.1. "in a component of a compound property" [cf. Section 2.1.3]
- [x] [4] Require `VERSION` to be "2.1" [cf. Section 2.6.6].
- [5] [6] n/a
- [x] [7] Support `CHARSET` type parameter [and default to `ASCII`, cf. 2.1.6].
- [ ] [8] Support non-significant WSP characters [i.e. HTAB, cf. 2.9].
- [x] [9] Don't require parameter name, if the value is unambiguous [e.g. for `TYPE=` and `ENCODING=`, cf. 2.1.2].
- [ ] [10] Require `TYPE` of  `LOGO` [cf. 2.5.3.1], `PHOTO` [cf. 2.2.3.1] and `SOUND` [cf. 2.6.3.1] as defined by the spec [e.g. `GIF`].
- [ ] [11] Parse inline binary content according to `ENCODING` parameter [and default to `7-bit`, cf. 2.1.5].
- [ ] [12] Do not support the `PCS` type parameter value for `TEL` [from vCard 3.0].
- [x] [13] Do not support `CATEGORIES`, `CLASS`, `NICKNAME`, `PRODID` and `SORT-STRING` [from vCard 3.0].
- [x] [14] Do not require `N` and `FN` [*Note*: `N` is only required for writers, not for readers, cf. 2.2.2].
- [14] *Note*: Despite what vCard 3.0 states,  `VERSION` is also required in vCard 2.1 [cf. [4] and 2.6.6].
- [x] [...] Use `COMMA` to separate `GEO` values [cf. 2.4.6 versus [`SEMICOLON` in vCard 3.0](https://tools.ietf.org/html/rfc2426#section-3.4.2)]
- [x] [MR] Add tests.
- [ ] [MR] Fix test coverage.


skjnldsv:
Awesome!! @caugner 
Let me know if I can be any help!

tasn:
Glad to see you are working on it, life saver!

I tried testing it, but am getting "unsupported vCard version" with a vCard 4.0. Taking a brief look at the code, it look like your switch statement only handles 3.0 and 2.1 (though it could be intentional in that specific spot).

caugner:
@tasn Can you share the vCard 4.0 which triggered the error? It sounds as if the vCard combines properties of vCard 3 and 4, because [this switch statement](https://github.com/mozilla-comm/ical.js/pull/389/files#diff-696ef1394d5d79ac879812cc401d58d1R299) should only be evaluated for vCards other than 4.0: `!(name === 'version' && value === '4.0')`

lib/ical/parse.js
```diff
@@ -228,6 +231,8 @@ ICAL.parse = (function() {
       );
     }
 
+    var debug = line.includes("37.386013");
```
caugner, line 234:
```suggestion
```

tasn:
Apologies for the noise!
I just checked following your comment and it seems that the failing one is a purposefully malformed entry I created. Interesting that none of the other clients (ez-vcard, python-vobject) fail when they get to it.

I guess you can ignore my comment. The entry for reference:

```
BEGIN:VCARD
PRODID;VALUE=TEXT:-//iCal.js EteSync Web
VERSION:4.0
UID:0561483d-251d-44e7-bb63-0cfd6285ffe3
FN:This contact checks broken vcard version
REV;VALUE=DATE-TIME:20190503T091907
IMPP;TYPE=jabber:jabber:
END:VCARD
```

stefan123t:
regarding [1] there are two old issues at nextcloud contacts & server with some example ENCODING=QUOTED-PRINTABLE characters of danish and german users in case you want to extend your test case vcard21.vcf

https://github.com/nextcloud/contacts/issues/584
https://github.com/nextcloud/server/issues/10653

tasn:
I guess this was accidentally closed because of the deletion of `master`?

NickCrews:
I am trying to revive this in #958 if anyone wants to help. Still quite rough and not ready for review, I will remove the WIP mark when it is ready.

PR #395, opened 2019-06-06
ICAL.Event duration should take timezones into account

georgehrke:
ICAL.Event has a getter for duration that either returns the duration property or calculates it from dtstart and dtend.

The calculation did not take timezones into account. Since RFC 5545 allows to define different timezones on DTSTART and DTEND, this behaviour is not appropriate.
Start: 01.01.2012 10:20 NYC
End: 01.01.2012 12:50 L.A.

previous duration was 2 1/2 hours,
when applying the fix the duration is correct: 5 1/2 hours

### DIFF
--- a/lib/ical/event.js
+++ b/lib/ical/event.js
@@ -391,7 +391,7 @@ ICAL.Event = (function() {
     get duration() {
       var duration = this._firstProp('duration');
       if (!duration) {
-        return this.endDate.subtractDate(this.startDate);
+        return this.endDate.subtractDateTz(this.startDate);
       }
       return duration;
     },
--- a/test/event_test.js
+++ b/test/event_test.js
@@ -2,7 +2,7 @@ suite('ICAL.Event', function() {
 
 
   var testTzid = 'America/New_York';
-  testSupport.useTimezones(testTzid, 'America/Denver');
+  testSupport.useTimezones(testTzid, 'America/Denver', 'America/Los_Angeles');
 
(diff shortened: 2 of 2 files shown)


georgehrke:
There is a similar bug in https://github.com/mozilla-comm/ical.js/blob/62f873410ca0ca3428af47f0e3137358e7b81c5e/lib/ical/period.js#L111

Although in that case subtractDate and subtractDateTz won't make a difference, since start and end of a period are always given in UTC.

georgehrke:
Weird, I called `grunt package`. Not sure why the build fails 🤔

georgehrke:
> since start and end of a period are always given in UTC.

Actually [RFC 5545 3.3.9](https://tools.ietf.org/html/rfc5545#section-3.3.9) does not say anything about start and end having to be in UTC. Only [RFC 5545 3.8.2.6](https://tools.ietf.org/html/rfc5545#section-3.8.2.6) mentions that the time value must be in UTC.

There is also [no errata](https://www.rfc-editor.org/errata_search.php?rfc=5545) about this.

kewisch:
Thanks for the patch! I'm taking care of the conflicts.

To have timezones there needs to be a TZID parameter on the property. This is why it isn't explcitly mentioned in the spec for a value, because that has no indication of timezone other than differing between floating time (no suffix) and UTC (Z).

Let's keep the calculation in the period code as is.

PR #421, opened 2020-02-26
Fix timezone convertion in case a given timezone has a RRULE in the future and the start date in the past of the datetime in question

DeepDiver1975:
Please correct me if I'm wrong!

It looks like that the timezone conversion logic is not properly working for the given test timezone Europe/Berlin2 

The standard timezone does not apply but the daylight one does for the given datetime 2019-10-25T15:30:00Z

I'll try to come up with a fix. Please meanwhile let me know if the test assumption is correct. THX

```ics
BEGIN:VCALENDAR
PRODID:-//tzurl.org//NONSGML Olson 2012h//EN
VERSION:2.0
BEGIN:VTIMEZONE
TZID:Europe/Berlin2
BEGIN:STANDARD
DTSTART:20191027T030000
TZOFFSETFROM:+0200
TZOFFSETTO:+0100
TZNAME:CET
END:STANDARD
BEGIN:DAYLIGHT
DTSTART:20190331T020000
TZOFFSETFROM:+0100
TZOFFSETTO:+0200
RDATE:20200329T020000
TZNAME:CEST
END:DAYLIGHT
END:VTIMEZONE
END:VCALENDAR
```


DeepDiver1975:
Looks like only the RDATE of daylight is taken into account when computing the changes array:

Screenshot from 2020-02-26 11-31-52

DeepDiver1975:
Who can review this please? @kewisch maybe? THX a lot!

DeepDiver1975:
> Who can review this please? @kewisch maybe? THX a lot!

:wave: @kewisch

DeepDiver1975:
:zzz:

DeepDiver1975:
:wave: anybody out there?

kewisch:
Thanks for the patch, I'll see what I can do to get this in soon. I need to somehow fix my overflowing github notification emails folder.

kewisch:
Since you may have seen some other activity here recently, I'm acknowledging that I've seen this and still have not fixed my notifications. This change is slightly to large for me to make a call right now, though I really want to get some of these PRs merged. Hoping to try again when I have a larger chunk of time.

kewisch:
Alright.  I think I was mainly worried about this change because it affects timezone handling and that is kind of integral. It was/is a bit awkward to me to change an if/else into just doing both branches, but at the time I thought maybe I'm just not understanding the change. I should have simply asked. 

This may be because I let this fester for 5 years, but after re-applying this there are a bunch of tests failing with the timezones off. If I just apply your tests, then only the Europe/Berlin2 test fails.

I'm guessing you've long moved on, so I am going to close this PR for now. I don't feel great about how long it took and the outcome I've created and I apologize for that. Either way, I appreciate that you had taken the time to get this set up. In the unlikely event you are still inclined to pick this up (hopefully not with another 5 years of review time), please feel free to re-open or create a new one.

DeepDiver1975:
> I'm guessing you've long moved on,

After 5 years ... yes. :wave: So long ....

PR #438, opened 2020-04-19
recur_expansion.js: EXDATE can be DATE and DTSTART can be DATE-TIME

dilyanpalauzov:
When EXDATE is in DATE format and DTSTART is DATE-TIME, then to determine whether a ocurrence shall be excluded, the occurence shall be converted to DATE and then compared to EXDATE.

This applies also for the very first EXDATE, when it coincides with DTSTART.

### DIFF
--- a/lib/ical/recur_expansion.js
+++ b/lib/ical/recur_expansion.js
@@ -44,7 +44,7 @@ ICAL.RecurExpansion = (function() {
    * });
    *
    * // remember there are infinite rules
-   * // so its a good idea to limit the scope
+   * // so it is a good idea to limit the scope
    * // of the iterations then resume later on.
    *
    * // next is always an ICAL.Time or null
--- a/lib/ical/recur_expansion.js
+++ b/lib/ical/recur_expansion.js
@@ -214,6 +214,21 @@ ICAL.RecurExpansion = (function() {
     },
 
     /**
+     * Compare two ICAL.Time objects.  When the second parameter is a timeless date
+     * and the first parameter is date-with-time, strip the time and compare only
+     * the days.
+     *
+     * @private
+     * @param {ICAL.Time} a   The one object to compare
+     * @param {ICAL.Time} b   The other object to compare
+     */
+    _compare_special: function(a, b) {
+        if (!a.isDate && b.isDate)
+	    return new ICAL.Time({year:a.year, month:a.month, day:a.day}).compare(b);
+        else return a.compare(b);
+    },
+
+    /**
      * Retrieve the next occurrence in the series.
      * @return {ICAL.Time}
      */
@@ -264,9 +279,10 @@ ICAL.RecurExpansion = (function() {
 
         // check the negative rules
         if (this.exDate) {
-          compare = this.exDate.compare(this.last);
+          //EXDATE can be in DATE format, but DTSTART is in DATE-TIME format
+          compare = this._compare_special(this.last, this.exDate);
 
-          if (compare < 0) {
+          if (compare > 0) {
             this._nextExDay();
           }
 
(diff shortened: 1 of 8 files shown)


tasn:
Thanks for the fix, but it doesn't seem to fix the issues in #458. Could you take a look? Maybe you missed something?
Thanks!

dilyanpalauzov:
No.  This is from April 2020 and the other is from November 2020.

tasn:
Yeah, I saw that, but they seem to be addressing the same underlying issue, so maybe the fix here is incomplete?
Thanks for your work btw!

dilyanpalauzov:
I patch locally ical.js to fix problems I have and submit the changes for inclusion upstream.  You can see under the current Pull Requests list, that I have done this more than once.  I have not experienced the problem you refer to.  The fixes I submit are not considered within reasonable time and essentially ignored (N.B. the last commit in the master branch is from me: I say this for completeness).  If I work on a problem I do not have, and submit patch upstream, for me it is not foreseeable, whether my work will eventually have impact on the common good, or will be just lost time.

tasn:
I think I didn't explain myself very well. I'm very appreciative of your fixes, and I realise you are scratching your own itches. I'm thankful for you making this library better, as I rely on it in my app!
I was just trying to point out that what this seems to fix doesn't fix the other issue which I thought was very similar and closely related. I wanted to point that out because: I obviously want it fix, but also because it may indicate your current fix is incomplete.

As for your PRs not being reviewed: yup, that's annoying. I feel you. :|

Thanks again for all your work!

lib/ical/recur_expansion.js
```diff
@@ -264,9 +279,10 @@ ICAL.RecurExpansion = (function() {
 
         // check the negative rules
         if (this.exDate) {
-          compare = this.exDate.compare(this.last);
+          //EXDATE can be in DATE format, but DTSTART is in DATE-TIME format
+          compare = this._compare_special(this.last, this.exDate);
 
-          if (compare < 0) {
+          if (compare > 0) {
```
kewisch, line 285: Can you explain why we need to switch the sign here?

lib/ical/recur_expansion.js
```diff
@@ -213,6 +213,21 @@ ICAL.RecurExpansion = (function() {
...
+    /**
+     * Compare two ICAL.Time objects.  When the second parameter is a timeless date
+     * and the first parameter is date-with-time, strip the time and compare only
+     * the days.
+     *
+     * @private
+     * @param {ICAL.Time} a   The one object to compare
+     * @param {ICAL.Time} b   The other object to compare
+     */
+    _compare_special: function(a, b) {
```
kewisch, line 225: If you don't mind maybe we can do some bikeshedding here. "Special" seems fairly general, not easy to understand what the function does just based on name. I was thinking `compareTimeless` but then again it doesn't always compare without time. Do you have a suggestion on how we can make this more clear?

kewisch (changes requested):
Would be great if you could add some tests for this case. I skipped reviewing the parts that belong to your other PRs, let's see if we can get them merged first :)

lib/ical/recur_expansion.js
dilyanpalauzov: I cannot explain this today.  When the case was opened last year, I was aware of the context.  Now I do not know and the only way to find out, is to dig in the code.  Anyway, two lines above the comparisson function is changed from compare=this.exDate.compare to compare=this._compare_special and using tests somehow, the sign change was found to be correct.

lib/ical/recur_expansion.js
dilyanpalauzov: I have no suggestion.  You can rename it as you want.

PR #439, opened 2020-04-25
helpers.foldline: do not insert spaces between the high and low UTF-16 surrogates

dilyanpalauzov:
https://tools.ietf.org/html/rfc5545#section-3.1 says:

   Note: It is possible for very simple implementations to generate
   improperly folded lines in the middle of a UTF-8 multi-octet
   sequence.  For this reason, implementations need to unfold lines
   in such a way to properly restore the original sequence.

ical.js used to fold in the middle of a UTF-8 multi-octet sequence.

### DIFF
--- a/lib/ical/helpers.js
+++ b/lib/ical/helpers.js
@@ -287,11 +287,22 @@ ICAL.helpers = {
    */
   foldline: function foldline(aLine) {
     var result = "";
-    var line = aLine || "";
-
+    var line = aLine || "", pos = 0, line_length = 0;
+    //pos counts position in line for the UTF-16 presentation
+    //line_length counts the bytes for the UTF-8 presentation
     while (line.length) {
-      result += ICAL.newLineChar + " " + line.substr(0, ICAL.foldLength);
-      line = line.substr(ICAL.foldLength);
+      var cp = line.codePointAt(pos);
+      if (cp < 128) ++line_length;
+      else if (cp < 2048) line_length += 2;//needs 2 UTF-8 bytes
+      else if (cp < 65536) line_length += 3;
+      else line_length += 4; //cp is less than 1114112
+      if (line_length < ICAL.foldLength + 1)
+        pos += cp > 65535 ? 2 : 1;
+      else {
+        result += ICAL.newLineChar + " " + line.substring(0, pos);
+        line = line.substring(pos);
+        pos = line_length = 0;
+      }
     }
     return result.substr(ICAL.newLineChar.length + 1);
   },
--- a/test/stringify_test.js
+++ b/test/stringify_test.js
@@ -113,6 +113,12 @@ suite('ICAL.stringify', function() {
       assert.equal(ICAL.stringify.property(subject.toJSON(), ICAL.design.icalendar, false), "DESCRIPTION:foo" + N + "bar");
       assert.equal(ICAL.stringify.property(subject.toJSON(), ICAL.design.icalendar, true), "DESCRIPTION:foobar");
 
+      var utf16_muscle = '\uD83D\uDCAA'; //in UTF-8 this is F0 DF 92 AA.  If space/new line is inserted between the surrogates, then the JS Engine substitutes each stand-alone surrogate with REPLACEMENT CHARACTER 0xEF 0xBF 0xBD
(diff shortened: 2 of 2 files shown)


dilyanpalauzov:
I was wrong with the above comment, here the updated one:

ical.js used to fold in the middle of a UTF-16 code point sequence.  To be precise, it inserted a space and a new line after the high UTF-16 surrogate and before the low UTF-16 surrogate.  This created invalid stand-alone surrogates which the JavaScript engine convers to something valid.

This is the problem, with Node.js:
```javascript
import {writeFileSync} from 'fs';

const a = '\uD83D\uDCAA'  //this is 💪 'Flexed Biceps', UTF-8 encoding: 0xF0 0x9F 0x92 0xAA, HTML Entity: &#128170; &#x1F4AA; UTF-16 encoding: 0xD83D 0xDCAA
writeFileSync('a1', a, 'utf8')  //file contains the bytes f0 9f 92 aa
writeFileSync('a2', a.charAt(0) + a.charAt(1), 'utf8') //file contains the bytes f0 9f 92 aa = REPLACEMENT CHARACTER = �
writeFileSync('b', a.charAt(0) + ' ' + a.charAt(1), 'utf8') //file contains the bytes 0xEF 0xBF 0xBD 0x20 0xEF 0xBF 0xBD
```

so the result in b is valid UTF-8, but it has nothing in commont with the original text and it cannot be suspected, that anyform of reconstructing b  to a shall be tried.

This patch takes a full UTF-16 character from the input,
• calculates whether it takes one or two UTF-16 chars, and keeps in pos where the next full UTF-16 character starts,
• calculates for the full UTF-16 code point, whether it needs 1, 2, 3 or 4
 bytes to be presened in UTF-8, keeping in line_length the bytes for UTF-8 so far necessary,
 • splits the line, when the UTF-8 presentation exceeds ICAL.foldLength bytes

dilyanpalauzov:
Added two test cases to test/stringify_test.js:test('folding',… , that fail with the current code:
```javascript
//ICAL.foldLength = 15; 
//'💪'  is in UTF-16 '\uD83D\uDCAA' but in UTF-8 this is F0 DF 92 AA.  If space/new line is inserted between the surrogates, then the JS Engine substitutes each stand-alone surrogate with REPLACEMENT CHARACTER 0xEF 0xBF 0xBD
subject.setValue('💪');
assert.equal(ICAL.stringify.property(subject.toJSON(), ICAL.design.icalendar, false), "DESCRIPTION:" + N + '💪');//verify new line is after ':', as otherwise the whole line is longer than ICAL.foldLength
subject.setValue('aa💪💪a💪💪');
assert.equal(ICAL.stringify.property(subject.toJSON(), ICAL.design.icalendar, false), "DESCRIPTION:aa" + N + '💪💪a💪' + N + '💪');//verify that 💪 is moved as whole to a new line as it is 4 UTF-8 bytes
```

dilyanpalauzov:
Internet Explorer does not have String.prototype.codePointAt, which is used here.  The code can be rewritten without this method, but it is unclear on what JavaScript engines ical.js is supposed to run.

lib/ical/helpers.js
```diff
@@ -287,11 +287,22 @@ ICAL.helpers = {
...
-    var line = aLine || "";
-
+    var line = aLine || "", pos = 0, line_length = 0;
+    //pos counts position in line for the UTF-16 presentation
+    //line_length counts the bytes for the UTF-8 presentation
     while (line.length) {
-      result += ICAL.newLineChar + " " + line.substr(0, ICAL.foldLength);
-      line = line.substr(ICAL.foldLength);
+      var cp = line.codePointAt(pos);
+      if (cp < 128) ++line_length;
```
kewisch, line 295: Can you run the linter here? I believe I have it set to use {} and newlines for all if statements instead of keeping them on one line.

Asde from that this seems reasonable to me, given there are tests I'm good with merging this.

kewisch (changes requested):

lib/ical/helpers.js
dilyanpalauzov: The only problem eslint had, were missing terminating semi-colons.

dilyanpalauzov:
When will a new version be released?

PR #440, opened 2020-05-05
Better handling of property groups

### LINKED ISSUE #411: Better handling of vCard property groups
Property names can have a prefix indicating the group they are associated with. For example:
```
BEGIN:VCARD
VERSION:3.0
UID:4445-07DE0202-0206-0B34-FF16-00A2C
PRODID:-//Apple Inc.//iOS 11.2.6//EN
FN:Mr. Smith
item1.EMAIL;type=INTERNET,pref:main@email.com
EMAIL;type=INTERNET,WORK:work@email.com
END:VCARD
```

Here we have both `item1.EMAIL` and `EMAIL`. There are two problems with how ical.js handles these:
1. `comp.getFirstProperty('email')` (and other such functions) will only return the second email which is confusing and makes handling these much harder.
2. The field name is not correctly processed so a few things like field type will be incorrect.

sildur:
Property names can be prefixed with a group name. This handles these
cases so the field types are correctly set, and the data correctly
parsed.

Partially fixes mozilla-comm/ical.js#411 (second item)

### DIFF
--- a/lib/ical/parse.js
+++ b/lib/ical/parse.js
@@ -232,9 +232,17 @@ ICAL.parse = (function() {
     var multiValue = false;
     var structuredValue = false;
     var propertyDetails;
+    var ungroupedName;
 
-    if (name in state.designSet.property) {
-      propertyDetails = state.designSet.property[name];
+    // fetch the ungrouped part of the name
+    if (name.indexOf('.') !== -1) {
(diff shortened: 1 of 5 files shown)


kewisch:
IIUC #442 supercedes this patch, so I am closing. Let me know if I got that wrong.

PR #442, opened 2020-05-13
Handle property groups round trip vCard <-> jCard according to spec

### LINKED ISSUE #411: Better handling of vCard property groups

castaway:
Based on: https://github.com/mozilla-comm/ical.js/pull/440

Updated to use group conversion rules: https://tools.ietf.org/html/rfc7095#section-3.3.1.2.1

Fix #411

### DIFF
--- a/lib/ical/parse.js
+++ b/lib/ical/parse.js
@@ -232,9 +232,17 @@ ICAL.parse = (function() {
     var multiValue = false;
     var structuredValue = false;
     var propertyDetails;
+    var ungroupedName;
 
-    if (name in state.designSet.property) {
-      propertyDetails = state.designSet.property[name];
+    // fetch the ungrouped part of the name
+    if (name.indexOf('.') !== -1) {
+      ungroupedName = name.split('.')[1];
+    } else {
+      ungroupedName = name;
+    }
+
+    if (ungroupedName in state.designSet.property) {
+      propertyDetails = state.designSet.property[ungroupedName];
 
       if ('multiValue' in propertyDetails) {
         multiValue = propertyDetails.multiValue;
--- a/lib/ical/parse.js
+++ b/lib/ical/parse.js
@@ -232,11 +232,14 @@ ICAL.parse = (function() {
     var multiValue = false;
     var structuredValue = false;
     var propertyDetails;
+    var splitName;
     var ungroupedName;
 
     // fetch the ungrouped part of the name
     if (name.indexOf('.') !== -1) {
-      ungroupedName = name.split('.')[1];
+      splitName = name.split('.');
+      params['group'] = splitName[0];
+      ungroupedName = splitName[1];
     } else {
       ungroupedName = name;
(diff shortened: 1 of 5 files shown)


lib/ical/stringify.js
```diff
@@ -108,7 +108,15 @@ ICAL.stringify = (function() {
...
-    var line = name;
+    var groupName = params['group'];
+    var line;
+    if (groupName) {
+      line = groupName + "." + name;
+    } else {
+      line = name;
+    }
+
+    delete params.group;
```
darktrojan, line 119: This shouldn't be deleted as the jCal is then changed for future operations. Instead the loop below should do nothing if `paramName` is `group`.

lib/ical/stringify.js
```diff
@@ -108,7 +108,15 @@ ICAL.stringify = (function() {
     var jsName = property[0];
     var params = property[1];
 
-    var line = name;
+    var groupName = params['group'];
+    var line;
+    if (groupName) {
+      line = groupName + "." + name;
```
darktrojan, line 114: `groupName` should be made upper-case, since `name` is upper-case.

lib/ical/stringify.js
```diff
@@ -108,7 +108,15 @@ ICAL.stringify = (function() {
     var jsName = property[0];
     var params = property[1];
 
-    var line = name;
+    var groupName = params['group'];
```
kewisch, line 111: This stringifies both jCard and jCal. The group param does not work that way in jCal though. We may need to have a hook from the designSet that allows augmenting the property name.

lib/ical/parse.js
```diff
@@ -232,9 +232,20 @@ ICAL.parse = (function() {
     var multiValue = false;
     var structuredValue = false;
     var propertyDetails;
+    var splitName;
+    var ungroupedName;
+
+    // fetch the ungrouped part of the name
+    if (name.indexOf('.') !== -1) {
```
kewisch, line 239: Likewise to comment below, we need to differentiate iCalendar vs vCard here by using the designSet.

kewisch:
@darktrojan Is this fixed by #512 ?

darktrojan:
Yes.

PR #450, opened 2020-07-14
Detect type="module"

### LINKED ISSUE #449: cannot be used in type="module" mode
Hi,

I'm using [microbundle](https://github.com/developit/microbundle) to include "ical.js". It's `modern` and `module` options allow to include the bundled js as a `<script type="module" src="bundle.js"></script>`

However, when I start my application, I get the following error:

```js
Uncaught ReferenceError: assignment to undeclared variable ICAL
```

this is because `ICAL` is never officially initialized. 

https://github.com/mozilla-comm/ical.js/blob/dcfe5a67527e73f54db21fbc31f23670a29704b8/lib/ical/helpers.js#L11

Now, in a normal environment `<script src="bundle.js"></script>`, this wouldn't be a problem as strict mode isn't enabled by default. For Javascript modules though, [strict mode is always on](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode):

> ECMAScript 2015 introduced JavaScript modules and therefore a 3rd way to enter strict mode.  The entire contents of JavaScript modules are automatically in strict mode, with no statement needed to initiate it.

I'm not sure how to handle this exactly, but a naive guess would have been to do further feature-detection. E.g. of ES6 modules. [SO has an idea.](https://stackoverflow.com/questions/27922232/how-to-feature-detect-es6-modules)

TimDaub:
Fixes #449 

This is merely an attempt to fix the above mentioned issue. The problem with this fix this that so far it only detects if the Javascript modules feature is available in the browser. However, a developer could add a regular `<script>` tag in a browser that supports modules. Then this statement would still run.

I don't see it much of a problem though, as the added statement runs in addition to the statements below, fixing it for the problems described in the linked issue #449

### DIFF
--- a/lib/ical/helpers.js
+++ b/lib/ical/helpers.js
@@ -4,10 +4,10 @@
  * Portions Copyright (C) Philipp Kewisch, 2011-2015 */
 
 
-if ('noModule' in HTMLScriptElement.prototype) {
-  // In case ical.js is executed in a JavaScript module, strict mode is enabled
-  // and requires a global variable to be assigned explicitly to e.g. the window
-  // object.
+if (typeof window !== 'undefined' && 'noModule' in HTMLScriptElement.prototype) {
+  // In case ical.js is executed in the browser and a JavaScript module, strict
+  // mode is enabled and requires a global variable to be assigned explicitly
+  // to e.g. the window object.
   window.ICAL = {};
 }
 
--- a/lib/ical/helpers.js
+++ b/lib/ical/helpers.js
@@ -4,6 +4,13 @@
  * Portions Copyright (C) Philipp Kewisch, 2011-2015 */
 
 
+if ('noModule' in HTMLScriptElement.prototype) {
+  // In case ical.js is executed in a JavaScript module, strict mode is enabled
+  // and requires a global variable to be assigned explicitly to e.g. the window
+  // object.
+  window.ICAL = {};
+}
+
 /* istanbul ignore next */
 /* jshint ignore:start */
 if (typeof module === 'object') {


kewisch:
This seems like a viable quick fix. What I'd like to aim for is modernizing ical.js to use es6 modules instead of trying to cram everything onto a global variable, but still keep the `ICAL` prefix on the outside.

Can you add a comment inside the if block what this is for? Also, you may need to adjust the jshint and istanbul comments before so they apply to the correct blocks.

TimDaub:
> What I'd like to aim for is modernizing ical.js to use es6 modules

👍  Godspeed! Looking forward to those changes.

> Can you add a comment inside the if block what this is for? 

Left a comment.

> Also, you may need to adjust the jshint and istanbul comments before so they apply to the correct blocks.

Adjusted the comments.

TimDaub:
bump.

TimDaub:
Update: I noticed that when executing `ical.js` in nodejs, `HTMLScriptElement` isn't available from the global space and so
the ical.js wouldn't be executable.

So in fb1e238, I'm now making sure that `window` is available and only then I'm performing the module check.

TimDaub:
👋🏼 Hey,

this PR has now been open for half a year. I prefer to use the actual ical.js in my application and not continue to rely on my personal fork. So can you please come to a decision about merging this PR? Thanks!

Best,
Tim

kewisch:
I appreciate the work you've done here, even if the timeline for responding may not suggest so. I'm going to add your case to #470 and get it fixed there, let me know if you can think of a case where window is not defined but HTMLScriptElement is.

PR #457, opened 2020-10-28
event.isRecurring() does not repeat with RECURRENCE-ID without RANGE

dilyanpalauzov:
Evolution creates instances of events, which contain RRULE and RECURRENCE-ID without RANGE parameter.  Since the RANGE is not THISANDFUTURE, but absent, such events do not recur.

https://gitlab.gnome.org/GNOME/evolution/-/issues/1180

Currenlty isRecurring() returns under these conditions TRUE, instead of FALSE.

### DIFF
--- a/lib/ical/event.js
+++ b/lib/ical/event.js
@@ -286,13 +286,18 @@ class Event {
   }
 
   /**
-   * Checks if the event is recurring
+   * Checks if the event is recurring, thus has an RRULE or RDATE property, or a
+   * RECURRENCE-ID with the respective RANGE parameter.
    *
    * @return {Boolean}        True, if event is recurring
    */
   isRecurring() {
-    let comp = this.component;
-    return comp.hasProperty('rrule') || comp.hasProperty('rdate');
+    let comp = this.component, recurrenceId;
+    if (comp.hasProperty('rrule') || comp.hasProperty('rdate')) {
+      recurrenceId = comp.getFirstProperty('recurrence-id');
+      return !recurrenceId || !!recurrenceId.getParameter('range');
+    }
+    return false;
   }
 
   /**


MartenBE:
This bug also causes problems in the Nextcloud Calendar app when subscribing to an Outlook ics. Is there any update if this can be merged?

https://github.com/nextcloud/calendar/issues/2684

I've also noticed the following:

> Some platforms, notably Outlook and Lotus Notes do strange things with this, including setting DTSTART and RECURRENCE-ID to be the same. The UID still refers to the parent, however, so we should rely on UID, and should just ignore DTSTART for occurrences.

https://wiki.mozilla.org/Calendar:Recurrence_and_Exceptions
https://docs.microsoft.com/en-us/openspecs/exchange_standards/ms-stanoical/1c70d4c6-ba39-4965-8090-3020358bae34

lib/ical/event.js
```diff
@@ -283,13 +283,18 @@ ICAL.Event = (function() {
     },
 
     /**
-     * Checks if the event is recurring
+     * Checks if the event is recurring.  The event is not recurring if it has RECURRENCE-ID property without RANGE parameter.
```
kewisch:
```suggestion
     * Checks if the event is recurring. The event must have an RRULE or RDATE property, or a
     * RECURRENCE-ID with the respective RANGE parameter.
```

kewisch (changes requested):
I know it has been a while, but any chance you could add a test for this?

dilyanpalauzov:
I do not understand the test framework and learning it will cost significant amount of time, to add a simple test.  So, my answer is - no.

lib/ical/event.js
dilyanpalauzov: I have no objection to any wording of thr comments.  The last sentence should end with ' in order to be recurring’.

dilyanpalauzov:
Now updated to ES6.

PR #470, opened 2021-06-11
Add declaration of ICAL var

mcfedr:
This allows the module to not crash when used from nodejs

```
ReferenceError: ICAL is not defined
```

This shouldnt cause any problems with usage in the browser

Similar to

https://github.com/mozilla-comm/ical.js/issues/449
https://github.com/mozilla-comm/ical.js/issues/329

### DIFF
--- a/lib/ical/helpers.js
+++ b/lib/ical/helpers.js
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  * Portions Copyright (C) Philipp Kewisch, 2011-2015 */
 
-
+var ICAL;
 /* istanbul ignore next */
 /* jshint ignore:start */
 if (typeof module === 'object') {
--- a/package.json
+++ b/package.json
@@ -54,7 +54,8 @@
     "node": ">=10"
   },
   "scripts": {
-    "test": "grunt test-node"
+    "test": "grunt test-node",
+    "prepare": "grunt"
   },
   "files": [
     "build/ical.js",


package.json
```diff
@@ -54,7 +54,8 @@
     "node": ">=10"
   },
   "scripts": {
-    "test": "grunt test-node"
+    "test": "grunt test-node",
+    "prepare": "grunt"
```
kewisch, line 58: Can you explain this change?

lib/ical/helpers.js
```diff
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  * Portions Copyright (C) Philipp Kewisch, 2011-2015 */
 
-
+var ICAL;
 /* istanbul ignore next */
 /* jshint ignore:start */
 if (typeof module === 'object') {
```
kewisch, line 9: I'm thinking something like this would be a good start:

```javascript
/* istanbul ignore next */
/* jshint ignore:start */
if (typeof module === 'object') {
  // CommonJS, where exports may be different each time.
  var ICAL = module.exports;
} else if (typeof this !== "undefined" && typeof ICAL !== 'object') {/* istanbul ignore next */
  /** @ignore */
  this.ICAL = {};
} else if (typeof ICAL !== 'object') {
  var ICAL = {};
}
/* jshint ignore:end */
```

What I'm still not sure about is how to make things work in native es6 modules. To actually export `ICAL`, we need `export default ICAL`, but that again doesn't work when it is loaded from a non-module. I don't want to use eval, I guess the only other option is to transpile?

kewisch (changes requested):

package.json
mcfedr: [prepare](https://docs.npmjs.com/cli/v7/using-npm/scripts#life-cycle-scripts) script is run by npm when installing from git sources as well as before npm publish - this means when install this fork without publishing to npm repo the code is built at install.

lib/ical/helpers.js
mcfedr: I'm not sure what the advantage of this is? in the `this` case setting `this.ICAL` is equivilent to just having a `var` ICAL as its the same scope - at least in the browser global context

es6 isnt a goal here - I think you are right, unlikely to work without some separate build just for it, I guess `export` will cause a syntax error in other contexts.

mcfedr: Also now I look - anyway that `var` in the `if` is effectively lifted to the scope of the file - having `var ICAL` in two parts of the `if` causes a warning of duplicated definition.

kewisch: I thought this would not be the case given it is in two different branches, only one can ever be called. Happy to move this up though. I believe the `this.ICAL` case was added for use in `Components.utils.import()` which I am going to be bold and drop now. Taking #450 into account as well, what do you think about this?

```javascript

/* jshint ignore:start */
var ICAL;
/* istanbul ignore next */
if (typeof module === 'object') {
  // CommonJS, where exports may be different each time.
  ICAL = module.exports;
} else if (typeof HTMLScriptElement !== 'undefined' && 'noModule' in HTMLScriptElement.prototype) {
  // Until we use ES6 exports, using <script type="module"> we define ICAL on the window global.
  window.ICAL = ICAL = {};
} else if (typeof ICAL !== 'object') {
  ICAL = {};
}
/* jshint ignore:end */
```

mcfedr: The check for nomodule - basically whats the point? when we have a top level `var` - in a <script src=ical> that var will be exactly the same result as assigning to window.ICAl

kewisch: This was to cope with <script type=module> as in #450 . The global scope is not the window iirc, and until we officially support modules it would help make ICAL global.

alexhaller:
Hi, any updates on this topic?

kewisch:
Pushed these as eee72a7e46f5635125c48c44e82195de9eb6c943 and 9300908b33e41d25981fb73bac273509f84e3b8f

PR #475, opened 2021-07-04
Get and set color property in event.js

materpillar:
[RFC-7986](https://icalendar.org/RFC-Specifications/iCalendar-RFC-7986/) specifies new properties for events. One of those new properties is `color`.
Adding a setter and getter for this property can nicely improve display of ical resources.

### My use case
My sport club uses Nextcloud to edit the public calendars that are shown on the website of the sport club. There is a color coding in place for certain events (different teams of the club). The color coding helps both the club's board as well as normal members to identify teams.
The Nextcloud Calendar app uses the [Fullcalendar.io](https://fullcalendar.io/) library. Since v5.5.0, FullCalendar supports ical feed display, which is great for displaying public calendars on websites. In this particular use case, the usability is greatly improved when the display on the website mirrors exactly the display in Nextcloud. Currently, ical.js is not handling the color property, so FullCalendar can not take over the color property of an event.

This pull requests adds setter and getter for the color property.
I was not yet able to add tests for this property. It's basicallly only an addition of 6 lines of code, which makes me hope that it's not too big of a thing.

### Additional use case
ical.js is intended to replace Thunderbird's calendar (Lighting) dependency on libical, as lined out in the roadmap [here](https://developer.thunderbird.net/planning/roadmap#libical-removal-tb-91). If this plan/project is continued (let's hope so), the addition of the color property also comes in very handy for Thunderbird's calendar, which could include a function to select, whether an event is colored as defined in the ical resource, or colored by an overwriting user set color.


lib/ical/event.js
```diff
@@ -452,6 +452,18 @@ ICAL.Event = (function() {
       this._setProp('description', value);
     },
 
+    /**
+     * The event color.
+     * @type {String}
```
kewisch: Since this is not an rfc5545 property let's mention where it comes from.

```suggestion
     * The event color from [rfc7986](https://datatracker.ietf.org/doc/html/rfc7986)
     * @type {String}
```

kewisch (approved):

kewisch (changes requested):
Can you also add a test [here](https://github.com/mozilla-comm/ical.js/blob/master/test/event_test.js#L759) ?

materpillar:
I am not sure if adding `testProperty('color', 'turquoise');` is sufficient in the mentioned line.
Is there no change to one of the ics files required?
(it's my first time working on a javascript project)

kewisch:
From a brief look the ics file to add to is https://github.com/mozilla-comm/ical.js/blob/master/samples/recur_instances.ics

You can run the tests to find out by cloning the repo, `npm install` and then `grunt test-node:unit`

materpillar:
Latest commit added color property to the ics file. Tests seem to be fine, but I obtained also the same result before the latest commit.. So not exactly sure what is being tested. An empty color property apparently does also not fail the test..

```
remaining properties
      ✓ #location
      ✓ #summary
      ✓ #description
      ✓ #organizer
      ✓ #uid
      ✓ #sequence
      **✓ #color**
      ✓ #duration
      ✓ #attendees
      ✓ #recurrenceId
```

kewisch:
Pushed this as 1fca94408148d75ffcbb666814890651da9f1c67 with the tests fixed. You'll want to set a different value (I took red), and see if changing it works as expected.

PR #477, opened 2021-10-02
Declare the ICAL variable at the top level

daisylb:
This change allows ical.js to be imported as a module (which requires it to run in strict mode).

Without this, ical.js doesn't work with [Snowpack](https://www.snowpack.dev), which builds everything into JS Modules.


kewisch:
See #470. I need to just go ahead with the code there, I haven't gotten around to cleaning it up.

PR #507, opened 2022-04-15
parser._handleContentLine() parsing including ^ in the parameter value

dilyanpalauzov:
Prior this change the property `ATTENDEE=CN="Ж 4 <>'^':":mailto:1@example.org` is read as:
```javascript
  jCal: [
    'attendee',
    { cn: `Ж 4 <>'":` },
    'cal-address',
    '":mailto:1@example.org'
  ]
```
The value starts with quote, which is wrong.

I do not add a test, because, among other things, [I cannot run](https://github.com/mozilla-comm/ical.js/issues/506) the existing tests.

### DIFF
--- a/lib/ical/parse.js
+++ b/lib/ical/parse.js
@@ -417,7 +417,9 @@ parse._parseParameters = function(line, start, designSet) {
       value = line.slice(valuePos, nextPos);
     }
 
+    const length_before = value.length;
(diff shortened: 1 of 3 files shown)


dilyanpalauzov:
Now updated to the ES6-changes.

kewisch (approved):

PR #512, opened 2022-05-26
 Handle property groups round trip vCard <-> jCard according to spec

darktrojan:
Extends #442 with nits fixed and a test that ical is handled differently.


test/parser/grouped.ics
```diff
@@ -0,0 +1,3 @@
+BEGIN:VCALENDAR
+GROUP1.X-TEST;VALUE=TEXT:test!
```
darktrojan, line 2: This isn't valid iCal but it does prove the handling difference between iCal and vCard.

lib/ical/parse.js
```diff
@@ -232,9 +232,20 @@ ICAL.parse = (function() {
     var multiValue = false;
     var structuredValue = false;
     var propertyDetails;
+    var splitName;
+    var ungroupedName;
+
+    // fetch the ungrouped part of the name
+    if (name.indexOf('.') !== -1 && state.designSet.propertyGroups) {
```
kewisch: Minor nit, let's turn this around. Slight perf increase in skipping the indexOf for iCalendar:
```suggestion
    if (state.designSet.propertyGroups && name.indexOf('.') !== -1) {
```

lib/ical/stringify.js
```diff
@@ -108,10 +108,24 @@ ICAL.stringify = (function() {
...
+    var line;
+    if (groupName && designSet.propertyGroups) {
+      line = groupName.toUpperCase() + "." + name;
+    } else {
+      line = name;
+    }
 
     var paramName;
     for (paramName in params) {
+      if (paramName == 'group' && designSet.propertyGroups) {
```
kewisch:
```suggestion
      if (designSet.propertyGroups && paramName == 'group') {
```

kewisch (approved):
r+wc, feel free to squash and merge.

darktrojan:
> r+wc, feel free to squash and merge.

I don't have write access. I've also left the original two commits unsquashed because they're not mine and it doesn't feel right to hide somebody else's work.

Issue #515, opened 2022-06-22
Failure processing 2022-03-07T::


kaie:
A calendar found in the wild contains date time values of the following format: 2022-03-07T::

While this is clearly a bad format, it seems the calendar code attempts to be lenient, so it might be good to attempt to handle this, too.

Originally reported at https://bugzilla.mozilla.org/show_bug.cgi?id=1754412

kaie:
pull request: https://github.com/mozilla-comm/ical.js/pull/516

twalling:
I was seeing this error this week when trying to parse this ICS file `http://www.calendarlabs.com/ical-calendar/ics/76/US_Holidays.ics`

It's different than the ones I've seen from google which include a `VALUE=DATE` string with each `DTSTART` and `DTEND`. Whenever I got to the point of trying to access those properties and the value was being wrapped in an  `ICAL.Time` model I was seeing an error like you described above, `2021-01-01T::` in this case.

edd189:
> I was seeing this error this week when trying to parse this ICS file `http://www.calendarlabs.com/ical-calendar/ics/76/US_Holidays.ics`
> 
> It's different than the ones I've seen from google which include a `VALUE=DATE` string with each `DTSTART` and `DTEND`. Whenever I got to the point of trying to access those properties and the value was being wrapped in an `ICAL.Time` model I was seeing an error like you described above, `2021-01-01T::` in this case.

Reporting same issue here. Throwing an error when the DTSTART or DTEND does not contain a time component. This appears to be allowed by the ics standard: https://datatracker.ietf.org/doc/html/rfc5545#section-3.8.2

 Value Type:  The default value type is DATE-TIME.  The value type can be set to a DATE value type. ("DATE-TIME" / "DATE")

kewisch:
Let's continue in #620

PR #516, opened 2022-06-22
Fix Issue #515

kaie:
Fall back to date only parsing

### DIFF
--- a/lib/ical/time.js
+++ b/lib/ical/time.js
@@ -1150,11 +1150,24 @@
    * @return {ICAL.Time}            The date/time instance
    */
   ICAL.Time.fromString = function fromString(aValue, aProperty) {
-    if (aValue.length > 10) {
-      return ICAL.Time.fromDateTimeString(aValue, aProperty);
-    } else {
-      return ICAL.Time.fromDateString(aValue);
+    let useDateOnly = true;
+    let result;
+
+    try {
+      if (aValue.length > 10) {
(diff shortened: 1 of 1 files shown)


kaie:
@darktrojan could you please approve the workflows and have a look? Thanks

darktrojan:
I don't have those powers. @kewisch?

lib/ical/time.js
```diff
@@ -1150,11 +1150,24 @@
...
+
+    try {
+      if (aValue.length > 10) {
+        result = ICAL.Time.fromDateTimeString(aValue, aProperty);
+        // If we arrive here, parsing worked, we can skip the fallback.
+        useDateOnly = false;
+      }
+    } catch (e) {}
+
+    if (useDateOnly) {
```
kewisch, line 1164: Let's gate the leniency on `!ICAL.design.strict`

kewisch (changes requested):
A test would be marvelous. 

I'm not sure why the linters are failing, but at least on this branch we use var (almost) everywhere for backwards compat. I need to merge the ES6 branch before it diverts too much, here we can use all the modern things.

I suspect we passed the deadline you were looking to make? Apologies in advance. If not then I'd be fine skipping the test in order to push this out, as long as you can follow up with the test a bit later.

kaie:
abandoning per request from Lasana

PR #518, opened 2022-07-02
fix: update dependencies for fix security issues

workgroupengineering:


kewisch:
I've taken care of this (hopefully) in 7ba2cea. I couldn't get rid of all the security issues with npm outdated and npm audit, but given these are in dev dependencies I see less pressure to get them resolved.

PR #519, opened 2022-07-13
Allow date values for DTSTART and DTEND without VALUE parameter when …

metasansana:
…strict=false

### DIFF
--- a/lib/ical/design.js
+++ b/lib/ical/design.js
@@ -550,9 +550,23 @@ ICAL.design = (function() {
     "contact": DEFAULT_TYPE_TEXT,
     "created": DEFAULT_TYPE_DATETIME,
     "description": DEFAULT_TYPE_TEXT,
-    "dtend": DEFAULT_TYPE_DATETIME_DATE,
+    "dtend": {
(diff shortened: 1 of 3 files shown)


darktrojan:
This is unnecessary. The `date-time` type already defers to `date` when not in strict mode. See [here](https://github.com/mozilla-comm/ical.js/blob/caef8fadabca0240fa588fa46426d8ea200e503b/lib/ical/design.js#L354-L356) and [here](https://github.com/mozilla-comm/ical.js/blob/caef8fadabca0240fa588fa46426d8ea200e503b/lib/ical/design.js#L378-L379).

metasansana:
> This is unnecessary. The `date-time` type already defers to `date` when not in strict mode. See [here](https://github.com/mozilla-comm/ical.js/blob/caef8fadabca0240fa588fa46426d8ea200e503b/lib/ical/design.js#L354-L356) and [here](https://github.com/mozilla-comm/ical.js/blob/caef8fadabca0240fa588fa46426d8ea200e503b/lib/ical/design.js#L378-L379).

Agreed. Never mind this then.

PR #520, opened 2022-07-14
Upstream spelling corrections from comm-central

darktrojan:
Somebody ran codespell at some point, this patch upstreams the corrections so we can stop working around them.

### DIFF
--- a/lib/ical/stringify.js
+++ b/lib/ical/stringify.js
@@ -114,7 +114,7 @@ ICAL.stringify = (function() {
 
     var groupName = params.group;
     var line;
-    if (groupName && designSet.propertyGroups) {
+    if (designSet.propertyGroups && groupName) {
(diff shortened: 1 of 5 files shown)


lib/ical/stringify.js
```diff
@@ -114,7 +114,7 @@ ICAL.stringify = (function() {
 
     var groupName = params.group;
     var line;
-    if (groupName && designSet.propertyGroups) {
+    if (designSet.propertyGroups && groupName) {
```
darktrojan, line 117: This one's on me. I updated the two repos separately, and obviously missed this line.

dilyanpalauzov: How is this a “spelling correction”?  In case both sides of `&&` are false, the former code will execute faster.

PR #530, opened 2022-09-07
Fix recurrence iteration where there is a negative BYMONTHDAY rule.

darktrojan:
Originally [bug 1789362](https://bugzilla.mozilla.org/show_bug.cgi?id=1789362). Calculation of the initial recurrence date is wrong where `BYMONTHDAY` is negative.

For example, if the rule is `FREQ=MONTHLY;INTERVAL=3;BYMONTHDAY=-1` the event recurs at 2, 5, 8 etc. months instead of 3, 6, 9 etc.. I've figured out that this is because we're storing a negative day value in an `ICAL.Time`. When we attempt to read the value it gets normalised to the previous month because the day value is negative.

### DIFF
--- a/lib/ical/recur_iterator.js
+++ b/lib/ical/recur_iterator.js
@@ -214,7 +214,7 @@ ICAL.RecurIterator = (function() {
       this.last.second = this.setup_defaults("BYSECOND", "SECONDLY", this.dtstart.second);
       this.last.minute = this.setup_defaults("BYMINUTE", "MINUTELY", this.dtstart.minute);
       this.last.hour = this.setup_defaults("BYHOUR", "HOURLY", this.dtstart.hour);
-      this.last.day = this.setup_defaults("BYMONTHDAY", "DAILY", this.dtstart.day);
+      var dayOffset = this.last.day = this.setup_defaults("BYMONTHDAY", "DAILY", this.dtstart.day);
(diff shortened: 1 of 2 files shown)


test/recur_iterator_test.js
```diff
@@ -675,6 +676,28 @@ suite('recur_iterator', function() {
           '2015-03-31T08:00:00'
         ]
       });
+      testRRULE('FREQ=MONTHLY;INTERVAL=3;BYMONTHDAY=-1', {
```
leftmostcat, line 681: At a glance, it's not clear to me what this test is looking for. A comment would be good, or even a failure reason if `testRRULE()` provides for that.

lib/ical/recur_iterator.js
```diff
@@ -302,9 +302,13 @@ ICAL.RecurIterator = (function() {
         }
 
       } else if (this.has_by_data("BYMONTHDAY")) {
-        if (this.last.day < 0) {
+        // Attempting to access `this.last.day` will cause the date to be normalised and
```
leftmostcat, line 305: Appreciate the comments here, let me understand this review. :sweat_smile:

darktrojan:
I've added some comments to the individual tests. It's probably still as clear as mud.

I also moved the inner `if` condition out. Because it probably makes more sense there unless you happen to be reading it in the context of this patch. I should've done it originally but that's what hindsight is for.

leftmostcat (approved):

PR #418, opened 2020-02-08
Wrong first occurrence for Monthly rule with BYMONTHDAY in particular cases

mschroeder:
This is a editable copy of PR #232 as requested by @kewisch.

Original description:
This PR covers issues which occurs with monthly rules with BYMONTHDAY.
The first occurrence is not always calculated correctly when the BYMONTHDAY, as absolute value, is greater than the last day of the month where the event starts or the last day of the month where the occurrence should occur in case of INTERVAL in the rule.
Mainly the issues are described in Lightning's bugs:
https://bugzilla.mozilla.org/show_bug.cgi?id=1266797
https://bugzilla.mozilla.org/show_bug.cgi?id=386516#c11

PR #531, opened 2022-09-09
Fix recurrence iteration where there is a BYMONTHDAY rule greater than the number of days in the start month.

darktrojan:
This gets tricky. If the starting month doesn't have an occurrence (e.g. June has no 31st day), we need to find the next month that does. It's possible to have a sequence of 3 months without as 31st at an interval of 2 (February, April, June) so we might need to look for the next month three times.

(In the Gregorian calendar it IS possible to miss 7 occurrences if the interval is 12 and the start is a February late in a century, but that's absurd and I will be dead by then.)

### DIFF
--- a/lib/ical/recur_iterator.js
+++ b/lib/ical/recur_iterator.js
@@ -311,14 +311,29 @@ class RecurIterator {
       if (this.last.day > daysInMonth || this.last.day == 0) {
         throw new Error("Malformed values in BYDAY part");
       }
-    } else if (this.has_by_data("BYMONTHDAY") && dayOffset < 0) {
-      // Attempting to access `this.last.day` will cause the date to be normalised and
-      // not return a negative value. We keep the value in a separate variable instead.
+    } else if (this.has_by_data("BYMONTHDAY")) {
+      // Attempting to access `this.last.day` will cause the date to be normalised.
+      // So it will never be a negative value or more than the number of days in the month.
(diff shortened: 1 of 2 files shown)


darktrojan:
Apart from waiting for the ES6 work, is this otherwise okay? I'd like to get it heading towards shipping in Thunderbird.

dilyanpalauzov:
This is very similar to https://github.com/kewisch/ical.js/pull/418.  I have not gone in details, but for me it seems both changesets approach problems with BYMONTHDAY in a different ways.

mschroeder:
> This is very similar to #418. I have not gone in details, but for me it seems both changesets approach problems with BYMONTHDAY in a different ways.

It is and together with the the already merged PR #530, the PR #418 is no longer needed as its development stalled years ago.

dilyanpalauzov:
> … the PR #418 is no longer needed as its development stalled years ago.

Why don’t you close #418 then?

mschroeder:
> > … the PR #418 is no longer needed as its development stalled years ago.
> 
> Why don’t you close #418 then?

I do not like how you phrase things in your comments here in the last weeks. It sounds rather pushy and impolite.

I myself will close the other PR when I have time to cross check if the test cases there are worth keeping in addition to the ones added in the other two PRs. If anybody wants to close it earlier, he could do so.

leftmostcat (approved):

PR #534, opened 2022-09-16
recur_expansion.js: fix iterator with RDATE-only recurrence

dilyanpalauzov:
When iterating with
```javascript
  e is an event-component
  if (e.isRecurring() {
     let i = e.iterator();
     while (n = i.next()) {
       o = g.getOccurrenceDetails(obj)
       …
     }
  }
```
on the first iteration the DTSTART instance shall be returned.  It is returned, when there is RRULE.  In the lack of this change, on the first iteration, when RDATE is present without RRULE, the first RDATE instance is returned.

@darktrojan  how does Thunderbird handle VEVENTS with RDATE but without RRULE.  These are presented correct in the UI.  But the instances cannot be obtained by getting an iteratior of the event and then calling .next().  The reason is, that the very first iteration skips then the DTSTART-instance.  The DTSTART instance is not skipped by the iterator.next() when the VEVENT has RRULE.

### DIFF
--- a/lib/ical/recur_expansion.js
+++ b/lib/ical/recur_expansion.js
@@ -121,7 +121,7 @@ ICAL.RecurExpansion = (function() {
      * @type {Number}
      * @private
      */
-    ruleDateInc: 0,
+    ruleDateInc: -1,
 
     /**
      * Current position in exDates array
@@ -241,6 +241,14 @@ ICAL.RecurExpansion = (function() {
         // _after_ we choose a value this should be
         // the only spot where we need to worry about the
         // end of events.
+        if (this.ruleDateInc == -1) {
+            this._nextRuleDay();
+            if (!iter) {
+              // on the first iteration return DTSTART
+              return this.last;
+            }
+        }
+
         if (!next && !iter) {
           // there are no more iterators or rdates
           this.complete = true;
@@ -278,7 +286,7 @@ ICAL.RecurExpansion = (function() {
         }
 
         //XXX: The spec states that after we resolve the final
-        //     list of dates we execute exdate this seems somewhat counter
+        //     list of dates we execute exdate.  This seems somewhat counter
         //     intuitive to what I have seen most servers do so for now
         //     I exclude based on the original date not the one that may
         //     have been modified by the exception.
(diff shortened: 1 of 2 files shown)


darktrojan:
> @darktrojan how does Thunderbird handle VEVENTS with RDATE but without RRULE. These are presented correct in the UI. But the instances cannot be obtained by getting an iteratior of the event and then calling .next(). The reason is, that the very first iteration skips then the DTSTART-instance. The DTSTART instance is not skipped by the iterator.next() when the VEVENT has RRULE.

Thunderbird doesn't use this code for RDATEs, it maintains a list of them and does iteration itself. So I'm not familiar with how this bit works at all.

darktrojan (changes requested):
In my very limited understanding of how this particular bit works, this looks okay to me. But it needs updating for ES6.

kewisch:
@dilyanpalauzov looks like you are actively working on this one, anything I can do to help get this merged?

dilyanpalauzov:
This change for proposal is one and a half years old.  I do not know what is missing to get it merged.  I see now it contains a test, which is failing.  Sorry, I do not know anymore what was going on.

kewisch:
Ah no problem, I had seen a recently referenced commit which made me assume. If you do feel like picking this up again I'm happy to give reviews some priority.

lib/ical/recur_expansion.js
```diff
@@ -384,15 +392,14 @@ ICAL.RecurExpansion = (function() {
...
           this.ruleDateInc = 0;
           this.last = this.ruleDates[0].clone();
+          this.ruleDate = this.ruleDates[0];
         } else {
           this.ruleDateInc = ICAL.helpers.binsearchInsert(
             this.ruleDates,
             this.last,
             compareTime
-          );
+          ) - 1;
```
kewisch, line 401: It appears we're no longer setting ruleDate in this branch. Is that intentional?

test/recur_expansion_test.js
```diff
@@ -61,6 +61,29 @@ suite('recur_expansion', function() {
...
+      });
+      var expected = [
+        new Date(2012, 09, 02, 10),
+        new Date(2012, 10, 05, 10),
+        new Date(2012, 10, 10, 10),
+        new Date(2012, 10, 30, 10)
+      ], dates = [], next;
+
+      while (next = subject.next() ) {
+        dates.push(next.toJSDate());
```
kewisch, line 81: The reason the test is failing is because there is a PERIOD RDATE, which does not have a toJSDate.

kewisch:
I appreciate that you got this started, but it seems neither you nor me have time to work on this currently. I'm going to close the PR, but if you want to pick this up again any time in the future then you are very much welcome to.

PR #535, opened 2022-09-17
stringify: do not use unescapedIndexOf in property parameters

dilyanpalauzov:
… as \ is no escape character there.

When the propery parameter contains :, then it must be quoted, the colon cannot be escaped.

As the function stringify.propertyValue in fact stringifies property parameter values, it is renamed accordingly.


dilyanpalauzov:
Now updated to the ES6 changes.

dilyanpalauzov:
Superseded by https://github.com/kewisch/ical.js/pull/658.

PR #536, opened 2022-09-18
parse._parsedParameters: do not use helpers.unescapedIndexOf()

dilyanpalauzov:
iCalendar and vCard parameters do not use `\` as escape character in the property parameter values, hence `cn="\"X"` is not a quote X, but invalid. The proper escape character is `^` per RFC 6868.  This change replaces helpers.unescapedIndexOf(x,…) with x.indexOf(…) in order not to count `\` as escape character.


dilyanpalauzov:
Now updated to ES6.

kewisch (approved):

PR #538, opened 2022-09-19
Adapt to ES6 modules

kewisch:
Hey @darktrojan @leftmostcat @metasansana Here is my ES6 adaption of the library. I'd love to get this merged before we do any other major work, rebasing just the few changesets was a pain an prone to failure. I can help adapt pending PRs afterwards.

With the amount of changes you'll have to see how much review is useful, be sure to [ignore whitespaces](https://github.com/kewisch/ical.js/pull/538/files?w=1) while reviewing, it will save you about 5k lines and makes reading easier!


docs/validator.html
```diff
@@ -0,0 +1,104 @@
...
+        border-radius: 5px;
+        border: 1px solid black;
+        margin-top: 1ex;
+      }
+
+      hr {
+        border: 1px solid black;
+      }
+    </style>
+    <script type="text/javascript" src="https://unpkg.com/ical.js@1.5.0/build/ical.js"></script>
```
kewisch: This will need updating to 2.0.0 and modules once released.

lib/ical/binary.js
```diff
@@ -1,181 +1,174 @@
...
 /**
- * This symbol is further described later on
- * @ignore
+ * Represents the BINARY value type, which contains extra methods for encoding and decoding.
+ *
+ * @class
+ * @alias ICAL.Binary
  */
-ICAL.Binary = (function() {
+class Binary {
```
dilyanpalauzov, line 12: This can be changed to `export default class {`, skipping the name Binary and skipping the last line of the file (export default Binary).

Likewise in the other files.

rollup.config.js
```diff
@@ -0,0 +1,32 @@
+import { getBabelOutputPlugin } from '@rollup/plugin-babel';
+import { terser } from 'rollup-plugin-terser';
+
+export default {
+  input: 'lib/ical/module.js',
+  output: [
+    { file: 'dist/ical.js',  format: 'es', exports: "default" },
```
dilyanpalauzov, line 7: package.json contains `"main": "build/ical.js`, while the build scripts puts the compiled files under `dist/`.  I am sceptical whether this will work.

rollup.config.js
```diff
@@ -0,0 +1,32 @@
...
+      format: 'cjs',
+      plugins: [
+        getBabelOutputPlugin({ presets: ['@babel/preset-env'] })
+      ],
+    },
+    {
+      file: 'dist/ical.es5.min.js',
+      exports: "default",
+      format: 'cjs',
+      plugins: [
```
dilyanpalauzov, line 26: package.js (NodeJS) has a concept of “Conditional exports” — https://nodejs.org/api/packages.html#conditional-exports . It can utilize the right input for `import ` (ES6 modules) / `require()` (ES5 CJS), depending on how the user utilizes the library.  ical.js would benefit by offering “Conditional exports” of the ES6/ES5 output.

tools/scriptutils.js
```diff
@@ -0,0 +1,150 @@
...
+    }
+  }
+
+  lines.push("})();");
+
+  return lines.join("\n");
+}
+
+async function get_tzdb_version() {
+  let response = await fetch('https://www.iana.org/time-zones');
```
dilyanpalauzov, line 127: `async` functions sometimes delay the execution, without having any benefit.  The delay is caused by the need of the runtime to create many contexts for execution.  If here async does not provide any benefit, I am advice against using it.

package.json
```diff
@@ -20,30 +21,24 @@
...
     "karma-spec-reporter": "^0.0.34",
     "minami": "^1.2.3",
-    "mocha": "^10.0.0"
+    "mocha": "^10.0.0",
+    "node-fetch": "^3.2.0",
+    "rollup-plugin-terser": "^7.0.2",
+    "yauzl-promise": "^2.1.3"
   },
   "overrides": {
     "grunt-release": {
```
dilyanpalauzov: is this still necessary?

lib/ical/helpers.js
```diff
@@ -1,377 +1,314 @@
...
+      break;
+  }
 
-    ICAL.helpers.dumpn(arguments[0]);
-  },
+  if (cmpval < 0)
+    return mid; // insertion is displacing, so use mid outright.
+  else if (cmpval > 0)
+    return mid + 1;
+  else
```
dilyanpalauzov, line 183: All `else` after `return`s can be deleted.

lib/ical/helpers.js
```diff
@@ -1,377 +1,314 @@
...
+ * var child = ICAL.helpers.extend(parent, {
+ *   "bar": 123
+ * });
+ *
+ * @param {Object} source     The object to extend
+ * @param {Object} target     The object to extend with
+ * @return {Object}           Returns the target.
+ */
+export function extend(source, target) {
+  for (let key in source) {
```
dilyanpalauzov, line 307: How is this function `extend` better than `Object.assign()`?  Is ical.js supposed to work in ancient environments with defect `Object.assign()`?

lib/ical/binary.js
Zearin: Having a named class (instead of an anonymous export) is better for:

* Generating docs
* Logging
* Editor auto-completion / “IntelliSense”, etc.

lib/ical/helpers.js
Zearin: Alternatively:

```js
let  newObj = { ...oldObj }
```

lib/ical/binary.js
dilyanpalauzov: `export default class Binary {` would also work.

tools/scriptutils.js
Zearin: The problem is usually less to do with using `async` on the function, and more to do with knowing when to use—and more importantly, when **not** to use—the `await` keyword inside of an async function.

If you can accomplish the function’s goal with fewer uses of `await`, that lets the runtime keep moving at top speed.

That said, I don’t think this particular case is worth optimizing.  It’s just a simple timezone fetch.  I don’t see how it could impact performance in any human-detectable way.

lib/ical/binary.js
Zearin: 👍

lib/ical/helpers.js
kewisch: Certainly something we can remove. This was from times before Object.assign. The PR is already large though, lets take care in a follow up.

lib/ical/recur_iterator.js
```diff
@@ -32,1400 +61,1365 @@ ICAL.RecurIterator = (function() {
...
+      var pos = parseInt(matches[1] || 0, 10);
+      dow = Recur.icalDayToNumericDay(matches[2], aWeekStart);
+      return [pos, dow];
+    } else {
+      return [0, 0];
+    }
+  }
 
+  next_generic(aRuleType, aInterval, aDateAttr,
+                                      aFollowingAttr, aPreviousIncr) {
```
darktrojan: This white space is weird.

lib/ical/stringify.js
```diff
@@ -112,31 +112,22 @@ stringify.property = function(property, designSet, noFold) {
     line = name;
   }
 
-  var paramName;
-  for (paramName in params) {
-    if (designSet.propertyGroups && paramName == 'group') {
-      continue;
-    }
```
darktrojan: This has gone missing.

darktrojan (approved):
I haven't got the available time or sanity to go through this line by line, but I've scanned through it and pointed out the things that stood out. I didn't really take much notice of changes outside `lib` and `test` as others have already had a look and TBH they probably know more about those things than I do.

As long as the tests pass and you pick up the missing bit of `stringify.property` I found then I approve. It would be useful if you put this into Thunderbird (even hacking it in temporarily) and ran the tests there, including the address book tests that handle vCards.

lib/ical/binary.js
kewisch: I can't quite put my finger on it now, but I believe I had a specific reason for this. I believe it was connected to the "export default function' is another special case" section of article: https://jakearchibald.com/2021/export-default-thing-vs-thing-as-default/  and/or because there was a situation where it made sense to do it in the one or other file, and then it made sense to keep it that way for consistency. To get this in I'm going to keep it this way, but if someone wants to go through the exports and change them all, and it still works afterwords, I'm happy to take a PR.

rollup.config.js
kewisch: Good point. I don't think I've tested using this from node as a package extensively.

tools/scriptutils.js
kewisch: In this specific case, I don't think `fetch` has a mode where it can be called synchronously. Also agree it doesn't make much of a difference in this file based on its purpose, but good to think about in performance critical code.

package.json
kewisch: Good catch!

lib/ical/helpers.js
kewisch: Technically yes, aesthetically I like having them sometimes. I'd rather handle this via linter, but am personally not keen on making the change (the lack of brackets here triggers me more, but that should also be a linter change :)

This PR is already big enough, lets defer this.

lib/ical/stringify.js
kewisch: Added this back in, with a test.

kewisch:
> It would be useful if you put this into Thunderbird (even hacking it in temporarily) and ran the tests there, including the address book tests that handle vCards.

I was much hoping someone from the team could test this in Thunderbird, I don't have the setup currently so it would take me a bit to get back into the groove.

rollup.config.js
kewisch: Taken care of, and another reference to `build`. Thanks!

PR #550, opened 2022-11-08
Try again before avoiding death by recursion

darktrojan:
I'm not sure about this. I've made it so that multiple occurrences at the same time are ignored.

But that's not what libical does (at least the ancient version that ships with Thunderbird). That gives every occurrence, even if it's the same as the previous one. In the first test case I added, the expected output would be:

```
2022-01-31T08:00:00
2022-01-31T08:00:00 <-- same as before
2022-02-28T08:00:00
2022-03-31T08:00:00
2022-03-31T08:00:00 <-- same as before
2022-04-30T08:00:00
2022-05-31T08:00:00
2022-05-31T08:00:00 <-- same as before
2022-06-30T08:00:00
```

The same behaviour can be replicated here easily, just by pulling out the "death by recursion" error completely.

Does anyone have insight into which is the right thing to do here?

### DIFF
--- a/lib/ical/recur.js
+++ b/lib/ical/recur.js
@@ -103,12 +103,13 @@ class Recur {
 
       if (ucname in partDesign) {
         let partArr = value.split(',');
-        let partArrIdx = 0;
-        let partArrLen = partArr.length;
+        let partSet = new Set();
 
-        for (; partArrIdx < partArrLen; partArrIdx++) {
-          partArr[partArrIdx] = partDesign[ucname](partArr[partArrIdx]);
+        for (let part of partArr) {
+          partSet.add(partDesign[ucname](part));
         }
+        partArr = [...partSet];
+
         dict[name] = (partArr.length == 1 ? partArr[0] : partArr);
       } else if (ucname in optionDesign) {
         optionDesign[ucname](value, dict, fmtIcal);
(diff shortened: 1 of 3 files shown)


darktrojan:
Something else that occurred to me while looking at this: if the data isn't in order, e.g. `BYMONTHDAY=3,1,2` then the occurrences will not be in chronological order, they'll be in the same order as the rule, e.g. `20221103`, `20221101`, `20221102`. That may or may not be a problem in practice (I think Thunderbird always wants all occurrences in a range so the order doesn't matter) and libical does the same thing, but it doesn't seem right.

darktrojan:
In the interests of fixing bad bugs, I might just remove the error in Thunderbird for now, since that produces the same behaviour as we got with libical.

darktrojan:
> In the interests of fixing bad bugs, I might just remove the error in Thunderbird for now, since that produces the same behaviour as we got with libical.

I did end up [removing the error in Thunderbird](https://github.com/darktrojan/ical.js/commit/8d6c9047424fde8f78062a52da57e1a6fab8d2e7). We haven't shipped the changes in this patch.

kewisch:
Thanks for the update. How confident do you feel about these changes? Should we give it a try and wait for feedback?

kewisch:
@darktrojan checking in on this

darktrojan:
It's certainly an improvement to the current code, and I can't think of any other cases that might trigger the error. Let's ship it and find out what happens.

kewisch (approved):

Issue #532, opened 2022-09-14
Parsing vCards adds additional quotation marks in TYPE atributes


5ila5:
```JavaScript
var vCard = "BEGIN:VCARD\nVERSION:3.0\nPRODID:-//Apple Inc.//iOS 13.3//EN\nN:Doe;John;;;\nFN:John Doe\nEMAIL;type=INTERNET;type=HOME;type=pref:John_Doer@example.com\nTEL;type=CELL;type=VOICE;type=pref:+49 123 4567890\nitem1.ADR;type=HOME;type=pref:;;street 2A;MyCity;MyState;83465;Deutschl\n and\nBDAY;value=date:1900-01-01\nREV:2020-01-17T10:34:14Z\nX-IMAGEHASH:qYML1ZJqQnL6FyrmNWNsWw==\nEND:VCARD"


var jcalData = ICAL.parse(vCard);
vCard = new ICAL.Component(jcalData)
console.log(vCard.toString())
```
returns 
```vCard
BEGIN:VCARD
VERSION:3.0
PRODID;VALUE=TEXT:-//Apple Inc.//iOS 13.3//EN
N:Doe;John;;;
FN:John Doe
EMAIL;TYPE="INTERNET,HOME,pref":John_Doer@example.com
TEL;TYPE="CELL,VOICE,pref";VALUE=PHONE-NUMBER:+49 123 4567890
ITEM1.ADR;TYPE="HOME,pref":;;street 2A;MyCity;MyState;83465;Deutschland
BDAY;VALUE=DATE:19000101
REV;VALUE=DATE-TIME:20200117T103414Z
X-IMAGEHASH:qYML1ZJqQnL6FyrmNWNsWw==
END:VCARD
```

Which has quotation marks for the Type parameter of EMAIL an TEL
I encountered this because i wanted to diagnose an error with Nextcloud where CardDav would display additional quotation marks 
https://github.com/nextcloud/server/issues/25485
https://github.com/nextcloud/contacts/issues/2866

I had this issue only when importing where this library is Used:
https://github.com/nextcloud/contacts/blob/5fd5d688d93d65b1079906e92eb31da65917b98d/src/models/contact.js#L61-L72

Edited:
the above code is called in followed by
vCard.toString()
https://github.com/nextcloud/contacts/blob/5fd5d688d93d65b1079906e92eb31da65917b98d/src/store/addressbooks.js#L410-L424

darktrojan:
[...]

PR #555, opened 2022-11-08
Quote (or don't quote) property values individually before joining

darktrojan:
Instead of adding quotes to the delimiting comma and around the outside, quote the individual members separately if necessary. This prevents the whole string of multiple values being quoted because of the presence of a delimiting comma.

When quotes are required (multiValueSeparateDQuote is true) only URIs are valid, which will get quoted due to the colon. When not required only specific strings (WORK, VOICE, PARCEL, etc.) are valid, none of which need quotes.

### DIFF
--- a/lib/ical/stringify.js
+++ b/lib/ical/stringify.js
@@ -116,20 +116,23 @@ stringify.property = function(property, designSet, noFold) {
     if (designSet.propertyGroups && paramName == 'group') {
       continue;
     }
-    let multiValue = (paramName in designSet.param) && designSet.param[paramName].multiValue;
+
+    let paramDesign = designSet.param[paramName];
+    let multiValue = paramDesign && paramDesign.multiValue;
     if (multiValue && Array.isArray(value)) {
-      if (designSet.param[paramName].multiValueSeparateDQuote) {
-        multiValue = '"' + multiValue + '"';
-      }
-      value = value.map(stringify._rfc6868Unescape);
+      value = value.map(function(val) {
+        val = stringify._rfc6868Unescape(val);
+        val = stringify.paramPropertyValue(val, paramDesign.multiValueSeparateDQuote);
+        return val;
+      });
       value = stringify.multiValue(value, multiValue, "unknown", null, designSet);
     } else {
       value = stringify._rfc6868Unescape(value);
+      value = stringify.paramPropertyValue(value);
     }
 
 
-    line += ';' + paramName.toUpperCase();
-    line += '=' + stringify.propertyValue(value);
+    line += ';' + paramName.toUpperCase() + '=' + value;
   }
 
   if (property.length === 3) {
@@ -206,13 +209,14 @@ stringify.property = function(property, designSet, noFold) {
  * If any of the above are present the result is wrapped
  * in double quotes.
  *
- * @function ICAL.stringify.propertyValue
+ * @function ICAL.stringify.paramPropertyValue
  * @param {String} value      Raw property value
+ * @param {boolean} force     If value should be escaped even when unnecessary
  * @return {String}           Given or escaped value when needed
  */
-stringify.propertyValue = function(value) {
-
-  if ((unescapedIndexOf(value, ',') === -1) &&
+stringify.paramPropertyValue = function(value, force) {
+  if (!force &&
+      (unescapedIndexOf(value, ',') === -1) &&
       (unescapedIndexOf(value, ':') === -1) &&
       (unescapedIndexOf(value, ';') === -1)) {
 
--- a/lib/ical/design.js
+++ b/lib/ical/design.js
@@ -193,7 +193,7 @@ let commonValues = {
 
 let icalParams = {
   // Although the syntax is DQUOTE uri DQUOTE, I don't think we should
-  // enfoce anything aside from it being a valid content line.
+  // enforce anything aside from it being a valid content line.
   //
   // At least some params require - if multi values are used - DQUOTEs
   // for each of its values - e.g. delegated-from="uri1","uri2"
--- a/test/stringify_test.js
+++ b/test/stringify_test.js
@@ -80,6 +80,39 @@ suite('ICAL.stringify', function() {
       delete ICAL.design.defaultSet.property.custom;
     });
 
+    test('property with multiple parameter values', function() {
+      ICAL.design.defaultSet.property.custom = { defaultType: 'text' };
+      ICAL.design.defaultSet.param.type = { multiValue: ',' };
+      let subject = new ICAL.Property('custom');
+      subject.setParameter('type', ['ABC', 'XYZ']);
+      subject.setValue('some value');
+      assert.equal(subject.toICALString(), 'CUSTOM;TYPE=ABC,XYZ:some value');
+      delete ICAL.design.defaultSet.property.custom;
+      delete ICAL.design.defaultSet.param.type;
+    });
+
+    test('property with multiple parameter values which must be escaped', function() {
+      ICAL.design.defaultSet.property.custom = { defaultType: 'text' };
+      ICAL.design.defaultSet.param.type = { multiValue: ',' };
+      let subject = new ICAL.Property('custom');
+      subject.setParameter('type', ['ABC', '--"XYZ"--']);
+      subject.setValue('some value');
+      assert.equal(subject.toICALString(), "CUSTOM;TYPE=ABC,--^'XYZ^'--:some value");
+      delete ICAL.design.defaultSet.property.custom;
+      delete ICAL.design.defaultSet.param.type;
+    });
+
+    test('property with multiple parameter values with enabled quoting', function() {
+      ICAL.design.defaultSet.property.custom = { defaultType: 'text' };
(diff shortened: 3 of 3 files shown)


darktrojan:
Aha, I thought I had asked about this before. I did, in #532.

dilyanpalauzov:
In the provided test cases:
```
ADR;TYPE=dom,home,postal,parcel:;;123 Main Street;Any Town;CA;91921-1234
TEL;TYPE=home,voice:1234567
```
the property values are `;;123 Main Street;Any Town;CA;91921-1234` and `1234567`.  The property parameter values are `dom,home,postal,parcel` and `home,voice`.  Just for clarification, can you state again, if the current case is about property values or property parameter values, as these terms can be easily confused.

Concerning the stringification of property parameter values and avoiding unnecessary quotes, I have a different approach — https://github.com/kewisch/ical.js/pull/535, for a different test-case.  In `A;X=Y:V` it does the stringification of the property value parameters (Y) independent of the property-value (V) and thus irrespective on whether the value-type is multi-valued.  I think this is the right approach.

I understand this is an open-source voluntary run project and I get what I paid for.  Different users of ical.js submit corrections for inclusion upstream.  Including the corrections upstream takes sometimes very, very long time.  This results that different solutions to the same problem are submitted, while uploaded corrections are not merged upstream.

I will appreciate if some process is suggested, that makes merging submitted changes, which include test cases, upstream more fluent.  This will be in the common interest.

I value the time invested in maintaining the project and personally I am not willing to be its maintainer.  The only point I am raising is that it takes very long time between patch submission and its inclusion.  So any speedup will be more than welcomed.

dilyanpalauzov:
Ah, this is not about multivalued property values, but about multivalued parameter values.  Can you add a test/example, which produces this stringification result:

> TEL;TYPE=home,voice,x-a,"x-b,x-c":1234567

darktrojan:
Yes, the names are confusing. I'd have changed the commit message with my recent push if only Github had notified me about your comments.

> Ah, this is not about multivalued property values, but about multivalued parameter values. Can you add a test/example, which produces this stringification result:
> 
> > TEL;TYPE=home,voice,x-a,"x-b,x-c":1234567

I'm not sure if that's valid. It's a question I've been wrestling with since I first tried to fix this bug. It looks like it should be valid, but I'm yet to find any concrete answer one way or the other. Maybe I'll just assume it is valid and move on.

darktrojan:
There. Added the extra bit to the test, which unsurprisingly was already working. I also renamed the function in question to be less confusing.

dilyanpalauzov:
https://www.rfc-editor.org/rfc/rfc6350#section-3.3 defines

any-param  = (iana-token / x-name) "=" param-value *("," param-value)
param-value = *SAFE-CHAR / DQUOTE *QSAFE-CHAR DQUOTE
type-param = "TYPE=" type-value *("," type-value)
type-value = "work" / "home" / type-param-tel / type-param-related / iana-token / x-name
x-name = "x-" 1*(ALPHA / DIGIT / "-")
  ; Names that begin with "x-" or "X-" are
  ; reserved for experimental use, not intended for released
  ; products, or for use in bilateral agreements.


QSAFE-CHAR = WSP / "!" / %x23-7E / NON-ASCII
     ; Any character except CTLs, DQUOTE

SAFE-CHAR = WSP / "!" / %x23-39 / %x3C-7E / NON-ASCII
     ; Any character except CTLs, DQUOTE, ";", ":"
                                    x22   x3A   x3B

Comma (ascii code dec 44, %x2C) is SAFE-CHAR and QSAFE-CHARE, as it is within %x23-39 and within %x23-7E.

I came now the conclusion that x-b,x-c as in `TEL;TYPE=home,voice,x-a,\"x-b,x-c\":1234567\r\n` is not permitted in TYPE as separate value: x-b,x-c is not an x-name, since it contains comma.  The ABNF of type-parameter (TYPE=) does not permit a comma or a quote in its value, when the comma is not separator, it permits x-name.

I have created one more example:
```diff
diff --git a/test/stringify_test.js b/test/stringify_test.js
index bc375b6..1ebd7c4 100644
--- a/test/stringify_test.js
+++ b/test/stringify_test.js
@@ -193,6 +193,14 @@ suite('ICAL.stringify', function() {
             },
             "phone-number",
             "1234567"
+          ],
+          [
+            "x-a",
+            {
+                "x-B": "x-c^d,"
+            },
+            "text",
+            "Brum"
           ]
         ]
       ];
@@ -200,6 +208,7 @@ suite('ICAL.stringify', function() {
         "BEGIN:VCARD\r\n" +
         "ADR;TYPE=dom,home,postal,parcel:;;123 Main Street;Any Town;CA;91921-1234\r\n" +
         "TEL;TYPE=home,voice,x-a,\"x-b,x-c\":1234567\r\n" +
+        "X-A;X-B=x-c^^d,;VALUE=TEXT:Brum\r\n" +
         "END:VCARD";
       assert.equal(ICAL.stringify.component(subject), expected);
     });
```

This fails.  The produced output is   X-A;X-B="x-c^^d,";VALUE=TEXT:Brum  but there is no need to quote, since  `x-c^^d,` are all SAFE-CHAR*.

```
  931 passing (964ms)
  1 failing

  1) ICAL.stringify
       stringify component
         multiple types unquoted:

      AssertionError: expected 'BEGIN:VCARD\r\nADR;TYPE=dom,home,post…' to equal 'BEGIN:VCARD\r\nADR;TYPE=dom,home,post…'
      + expected - actual

       BEGIN:VCARD
       ADR;TYPE=dom,home,postal,parcel:;;123 Main Street;Any Town;CA;91921-1234
       TEL;TYPE=home,voice,x-a,"x-b,x-c":1234567
      -X-A;X-B="x-c^^d,";VALUE=TEXT:Brum
      +X-A;X-B=x-c^^d,;VALUE=TEXT:Brum
       END:VCARD
      
      at assert.equal (node_modules/chai/lib/chai/interface/assert.js:139:10)
```

dilyanpalauzov:
https://github.com/kewisch/ical.js/pull/460 is a different approach to the same problem.

darktrojan:
I took the tests from #460 since they're more comprehensive than mine, and fixed a failure they caught.

lib/ical/stringify.js
```diff
@@ -116,20 +116,23 @@ stringify.property = function(property, designSet, noFold) {
     if (designSet.propertyGroups && paramName == 'group') {
       continue;
     }
-    let multiValue = (paramName in designSet.param) && designSet.param[paramName].multiValue;
+
+    let paramDesign = designSet.param[paramName];
+    let multiValue = paramDesign && paramDesign.multiValue;
```
dilyanpalauzov, line 121:
```javascript
const multiValue = paramDesign?.multiValue;
```

dilyanpalauzov:
The backslash is not escaping character in property parameters.  When I modify one of the tests as:
```diff
diff --git a/test/stringify_test.js b/test/stringify_test.js
index 40337b8..800fe19 100644
--- a/test/stringify_test.js
+++ b/test/stringify_test.js
@@ -84,9 +84,9 @@ suite('ICAL.stringify', function() {
       ICAL.design.defaultSet.property.custom = { defaultType: 'text' };
       ICAL.design.defaultSet.param.type = { multiValue: ',' };
       let subject = new ICAL.Property('custom');
-      subject.setParameter('type', ['ABC', 'XYZ']);
+      subject.setParameter('type', ['\\:ABC', 'XYZ']);
       subject.setValue('some value');
-      assert.equal(subject.toICALString(), 'CUSTOM;TYPE=ABC,XYZ:some value');
+      assert.equal(subject.toICALString(), 'CUSTOM;TYPE="\\:ABC",XYZ:some value');
       delete ICAL.design.defaultSet.property.custom;
       delete ICAL.design.defaultSet.param.type;
     });
```
it fails:
```
  1) ICAL.stringify                                                                                                                     
       stringify property                                                                                                               
         property with multiple parameter values:                                                                                       
                                                                                                                                        
      AssertionError: expected 'CUSTOM;TYPE=\:ABC,XYZ:some value' to equal 'CUSTOM;TYPE="\:ABC",XYZ:some value'                         
      + expected - actual                                                                                                               
                                                                                                                                        
      -CUSTOM;TYPE=\:ABC,XYZ:some value                                                                                                 
      +CUSTOM;TYPE="\:ABC",XYZ:some value                                                                                               

```

`\:ABC` must be quoted, since \ is not an escaping character and the property parameter value contains colon `:`.

https://github.com/kewisch/ical.js/pull/535 suggests to use `value.indexOf()` in stringify.propertyParameterValue instead of `helpers.unescapedIndexOf()`.

darktrojan:
Rebased with no further changes.

FWIW, we've been shipping these changes in Thunderbird since November 2022. It may not be perfect but I think we should merge as-is and open follow-up issues for outstanding problems.

kewisch (approved):

kewisch:
Sounds fine to me, thank you!

lib/ical/stringify.js
dilyanpalauzov: For me it is unclear, why the above suggestion was skipped.

dilyanpalauzov:
As I mentioned above, this change is suboptimal.  https://github.com/kewisch/ical.js/pull/658 does further improve encoding (quoting) of property parameter values.

PR #579, opened 2023-03-07
Support use of timezones specified in ICS

leftmostcat:
There is currently no support in ical.js for using a timezone defined within an ICS file for datetimes within that same ICS file. Instead, we are reliant on the TZID used in the file matching one in our time zone database.

In order to prevent bloating the timezone service with one-off time zones or using conflicting definitions, this patch restricts time zones to a single component tree, storing them on the root component and retrieving them from there when referenced.


test/recur_expansion_test.js
```diff
@@ -37,8 +37,8 @@ suite('recur_expansion', function() {
   suite('initialization', function() {
     test('successful', function() {
       assert.deepEqual(
-        new Date(2012, 9, 2, 10),
```
leftmostcat, line 40: It was necessary to modify these tests due to previous behavior not accounting for time zones. When calling the `Date()` ctor with numerical parameters, those parameters are interpreted as being in the user's time zone; the tests passed previously because the time zone definition in the files was not used and none of the tests in this file registered the `America/Los_Angeles` time zone with the time zone service.

lib/ical/time.js
```diff
@@ -165,25 +165,58 @@ class Time {
...
+          const rootZones = root.getAllSubcomponents('vtimezone');
+          for (const rootZone of rootZones) {
+            if (rootZone.getFirstProperty('tzid').getFirstValue() === zoneId) {
+              zone = new Timezone({
+                component: rootZone,
+                tzid: zoneId,
+              });
+
+              break;
+            }
```
kewisch: It sounds like it might be a good idea to cache the timezones on the root component lazily.

kewisch (changes requested):
Looks good, with one comment. I imagine that in a large file we may have quite a few timezone lookups, caching on the root element avoids reinstanciating the `Timezone` each time.

lib/ical/time.js
leftmostcat: Addressed. I've built a time zone cache map on the `VCALENDAR` component and moved the hydration of zones to there.

kewisch (approved):

lib/ical/time.js
```diff
@@ -165,25 +165,45 @@ class Time {
...
+      zone = Timezone.utcTimezone;
     } else if (prop) {
-      zone = prop.getParameter('tzid');
+      zoneId = prop.getParameter('tzid');
+
+      if (prop.parent) {
+        if (prop.parent.name === 'standard' || prop.parent.name === 'daylight') {
+          // Per RFC 5545 3.8.2.4 and 3.8.2.2, start/end date-times within
+          // these components MUST be specified in local time.
+          zone = Timezone.floating;
```
alex-kinokon, line 179: Where is `Timezone.floating` defined?

leftmostcat: This may be an error on my part; it should probably be `Timezone.localTimezone`.

Issue #367, opened 2018-09-20
Generate TypeScript ical.d.ts from jsdoc and include in releases


knu73r1k:
Anybody which know if there exists a ical.d.ts file for TypeScript @types module?

kewisch:
I'm mostly worried about keeping this file up to date. @glixlur Did you write that by hand, or was it automatically generated?

knu73r1k:
Cool - tx @glixlur

sebius8780:
@kewisch Do you have any work around to import ical.js into angular 6 project ?
If not @glixlur it will be really cool if you could work it out with @kewisch ?

i am looking forward to see if you can pull it of, would be really nice to have this lib Ts compliant :)

ALGDB:
hi, how can I use @glixlur types?

ALGDB:
Hi Just to understand. 

in: 

`new Component(jCal, parent)` 

`parent` is optional or mandatory as stated in the [docs](http://mozilla-comm.github.io/ical.js/api/ICAL.Component.html)?

because, as far as I can see it seems optional [component.js](http://mozilla-comm.github.io/ical.js/api/component.js.html)
```
function Component(jCal, parent) { 
if (typeof(jCal) === 'string') {
 // jCal spec (name, properties, components)
 jCal = [jCal, [], []];
}
 // mostly for legacy reasons.
 this.jCal = jCal; 
this.parent = parent || null;
  }
```

sebius8780:
I made it work using undefined, you should be able to make it right within the d.ts file though.

tasn:
I have a manually created (though incomplete) typing file I created. I use it extensively in and add parts as I get to them https://github.com/etesync/etesync-web The actual file: https://github.com/etesync/etesync-web/blob/master/src/types/ical.js.d.ts (I updated it a bit more since, will push those local changes soon I hope).
Also, I created it manually. If there's a way to create it automatically (@proteria ?) that would be much better (as long as it respects optional parameters). :P

I can take a look into making it more complete and submit a patch if there's interest. @kewisch ?

Otherwise it can go to https://github.com/DefinitelyTyped/DefinitelyTyped but then you'd have even less control when it comes to keeping it up to date.

kewisch:
I'd prefer something that is auto-generated, I don't want to be updating this file manually or requiring it for people making pull requests.

tasn:
[...]

PR #597, opened 2023-06-23
Added TypeScript declaration file and configuration

Aryan-mor:
This pull request adds a TypeScript declaration file (ical.js.d.ts) for ical.js to enable TypeScript code to be used with ical.js. A tsconfig.json file has also been included to provide TypeScript configuration for the library.

The ical.js.d.ts file contains type definitions for the Component and Event classes and their associated methods, as well as the parse function.

In addition, the tsconfig.json file sets up the TypeScript compiler options required to transpile TypeScript files to JavaScript.

This change will make it easy to use ical.js with TypeScript and provide better type-checking and code completion in editors.

### DIFF
--- a/lib/ical.js.d.ts
+++ b/lib/ical.js.d.ts
@@ -1,0 +1,239 @@
+declare module 'src/interfaces/ical.js' {
+  function parse(input: string): any[];
+
+  export class helpers {
+    public static updateTimezones(vcal: Component): Component;
+  }
+
+  class Component {
+    public static fromString(str: string): Component;
+
+    public name: string;
+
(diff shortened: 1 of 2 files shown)


kewisch:
I'd prefer not to take a manual approach here, since we'd need to maintain two separate file signatures. I'm ok adding some automation that would generate the d.ts from jsdoc, as noted in https://github.com/kewisch/ical.js/issues/367#issuecomment-1051350661 . I appreciate you made the effort to send a PR, if you'd like to work out automation based on the jsdoc I'm all for it!

PR #613, opened 2023-08-17
Fixes #612 -  a bug while parsing vcalendar parameters.

### LINKED ISSUE #612: Bug in parameter parsing function "Empty parameter name"
The parseParameters function does not work correctly when there is a property with parameter whose value is double quoted and in the property value there is semicolon and equal sign..

PoC taken from real vcard:
```
    BEGIN:VCALENDAR
    BEGIN:VEVENT
    X-ALT-DESC;FMTTYPE="text/html":Here is HTML with signs like =;
    SUMMARY;LANGUAGE=el:IMPACTS EDIC - Video conference schedule
    END:VEVENT
    END:VCALENDAR`
```
Here the X-ALT-DESC is breaking the logic. You can see this in this fiddle: https://jsfiddle.net/6d0m9uaf/

The problem is in parseParameters function which for quoted parameters just checks if there is another parameter delimeter not checking if it is past the value delimeter, lines:
https://github.com/kewisch/ical.js/blob/35533497954b2c6e20a902789fe11e95740c3cf6/lib/ical/parse.js#L390-L393

I will make a PR with a fix.

br0xpl:
Fixes #612

### DIFF
--- a/lib/ical/parse.js
+++ b/lib/ical/parse.js
@@ -388,7 +388,9 @@ parse._parseParameters = function(line, start, designSet) {
       }
       value = line.slice(valuePos, pos);
       lastParam = unescapedIndexOf(line, PARAM_DELIMITER, pos);
-      if (lastParam === -1) {
+      let propValuePos = unescapedIndexOf(line, VALUE_DELIMITER, pos);
+      // if either no next parameter or delimeter in property value, let's stop here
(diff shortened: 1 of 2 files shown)


br0xpl:
@kewisch can you have a look?

kewisch:
Hey, sorry for the long darkness. This sounds reasonable. @br0xpl I know it has been a while, but would you be willing to add a test case for this?

kewisch (approved):

PR #621, opened 2023-09-14
Avoid infinite loop with invalid YEARLY recurrence rule

leftmostcat:
See https://phabricator.services.mozilla.com/D188118 and https://github.com/darktrojan/ical.js/pull/12.


lib/ical/recur_iterator.js
```diff
@@ -179,7 +179,14 @@ class RecurIterator {
...
 
     if (!this.initialized) {
-      this.init();
+      try {
+        this.init();
+      } catch (e) {
+        // Init may error if there are no possible recurrence instances from the
+        // rule, but we don't want to bubble this error up. Instead, we create
+        // an empty iterator.
+        this.completed = true;
```
kewisch: Would it make sense to use a custom error class instead and catch only that one? Otherwise it would hide other exceptions during development.

test/recur_test.js
```diff
@@ -26,56 +26,56 @@ suite('recur', function() {
...
       }
     }, 'For WEEKLY recurrences neither BYMONTHDAY nor BYYEARDAY may appear');
 
-    checkThrow({
+    checkNoInstance({
       freq: 'DAILY',
       parts: {
         BYYEARDAY: [200]
       }
     }, 'BYYEARDAY may only appear in YEARLY rules');
```
kewisch: Spec says that `The BYYEARDAY rule part MUST NOT be specified when the FREQ
      rule part is set to DAILY, WEEKLY, or MONTHLY.`, so this combination should normally be an error instead of just not producing instances.

I guess that brings up the bigger question of should we just ignore such errors and be lenient, or should we throw. What about using `ICAL.design.strict` to determine if it should throw or silently show no instances?

kewisch (changes requested):

lib/ical/recur_iterator.js
leftmostcat: You're thinking using [this pattern](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#conditional_catch_blocks)?

test/recur_test.js
leftmostcat: The robustness principle suggests to me that we should only error if we reach a point where we simply can't interpret the file any longer. I'd be okay with using `strict` as a guard here if that makes sense to you. Probably the _best_ thing to do in this instance is to favor the `FREQ` property and any invalid parts that follow are silently dropped, resulting in occurrences that follow whatever can be validly interpreted. Unfortunately, the spec provides rules for how to _create_ valid iCalendar objects, but no real guidance on how to deal with invalid ones.

lib/ical/recur_iterator.js
kewisch: Yes

test/recur_test.js
kewisch: Yeah, generally I'd like ical.js to be able to parse everything and be lenient. The strict mode was meant for the case where you might use ical.js as a validator and would want it to bail. Let's put this behind the strict mode check.

leftmostcat: I wasn't quite sure how to access whether we were in strict mode from the recurrence iter, so I've implemented the infinite loop prevention without modifying any throwing behavior. If that behavior is desire, it may be best to address that in a followup.

leftmostcat:
I've addressed the comments left. The timezones CI job seems to be failing due to something entirely unrelated that I don't understand.

lib/ical/recur_iterator.js
```diff
@@ -1360,4 +1384,11 @@ class RecurIterator {
     return result;
   }
 }
+
+class InvalidRecurrenceRuleError extends Error {
```
kewisch, line 1395: Can you add some jsdoc here?

test/recur_iterator_test.js
```diff
@@ -143,14 +143,21 @@ suite('recur_iterator', function() {
...
         assert.throws(function() {
           recur.iterator(start);
         });
         return;
       }
+
       let iterator = recur.iterator(start);
 
+      if (options.noInstance) {
+        assert.equal(iterator.next(), null);
```
kewisch, line 157: You could also assert that `completed` is set, while you are at it.

kewisch (approved):
r+wc, thank you!

PR #654, opened 2024-04-03
Simplify lazy initialisation of TimezoneService.

darktrojan:
It appears the current way of bootstrapping the core zones exists because TimezoneService's position among the modules means we can't call reset immediately.

Something seem to not work right and leads to Time.epochTime being in the floating time zone instead of UTC. At least when run in Thunderbird, I'm not sure if it always happens. My patch just creates the zones on first use.

### DIFF
--- a/lib/ical/timezone_service.js
+++ b/lib/ical/timezone_service.js
@@ -6,26 +6,7 @@
 import Timezone from "./timezone.js";
 import Component from "./component.js";
 
-let zones = {
-};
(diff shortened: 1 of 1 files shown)


darktrojan:
Oops, I failed at git.

leftmostcat (approved):
This looks sound to me and I think getting rid of side effects for getting one of the three special zones is a big improvement.

PR #658, opened 2024-04-09
stringify: do not use unescapedIndexOf in property parameters

dilyanpalauzov:
… as `\` is no escape character there.  When the propery parameter contains `:`, then it must be quoted, the colon cannot be escaped.

As the function stringify.propertyValue in fact stringifies property parameter values, it is renamed accordingly.

This supersedes https://github.com/kewisch/ical.js/pull/535 and updates https://github.com/kewisch/ical.js/pull/555.

Without these changes, the herein added test fails:
```
1) ICAL.stringify
     stringify property
       stringify property value containing "escaped" ; , ::

    AssertionError: expected 'ATTENDEE;CN=X\::mailto:id' to equal 'ATTENDEE;CN="X\:":mailto:id'
    + expected - actual

    -ATTENDEE;CN=X\::mailto:id
    +ATTENDEE;CN="X\:":mailto:id
```

### DIFF
--- a/lib/ical/stringify.js
+++ b/lib/ical/stringify.js
@@ -4,7 +4,7 @@
  * Portions Copyright (C) Philipp Kewisch */
 
 import design from "./design.js";
-import { foldline, unescapedIndexOf } from "./helpers.js";
+import { foldline } from "./helpers.js";
 
 const LINE_ENDING = '\r\n';
 const DEFAULT_VALUE_TYPE = 'unknown';
@@ -131,7 +131,6 @@ stringify.property = function(property, designSet, noFold) {
       value = stringify.paramPropertyValue(value);
     }
 
-
     line += ';' + paramName.toUpperCase() + '=' + value;
   }
 
@@ -216,9 +215,9 @@ stringify.property = function(property, designSet, noFold) {
  */
 stringify.paramPropertyValue = function(value, force) {
   if (!force &&
-      (unescapedIndexOf(value, ',') === -1) &&
-      (unescapedIndexOf(value, ':') === -1) &&
-      (unescapedIndexOf(value, ';') === -1)) {
+      (value.indexOf(',') === -1) &&
+      (value.indexOf(':') === -1) &&
+      (value.indexOf(';') === -1)) {
 
     return value;
   }
--- a/test/stringify_test.js
+++ b/test/stringify_test.js
@@ -113,6 +113,13 @@ suite('ICAL.stringify', function() {
       delete ICAL.design.defaultSet.param.type;
     });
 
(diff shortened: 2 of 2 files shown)


test/stringify_test.js
```diff
@@ -113,6 +113,13 @@ suite('ICAL.stringify', function() {
       delete ICAL.design.defaultSet.param.type;
     });
 
+    test('stringify property value containing "escaped" ; , :', function() {
```
mschroeder: The part `; , ` does not make sense to me in the description of the test.

dilyanpalauzov: The proposal to add this test is from 2022, if you consider https://github.com/kewisch/ical.js/pull/535 . It is unclear to me, if any changes I do will be merged in reasonable time, and this is one of the reasons I prefer not to change anything further here.

In RFC 5545 the definitions are:
```
param-value   = paramtext / quoted-string
paramtext     = *SAFE-CHAR
SAFE-CHAR     = WSP / %x21 / %x23-2B / %x2D-39 / %x3C-7E
    / NON-US-ASCII
; Any character except CONTROL, DQUOTE, ";", ":", ","
quoted-string = DQUOTE *QSAFE-CHAR DQUOTE
```
In general, if a parameter value contains ";", ":" or "," then it must be quoted.  In the concrete case, the test does not verify all three cases.


Now, how is the test exactly called is a nuance.  It is a neglectable problem compared to the fact that this change is pending since 2022.  Once some expert decides to merge this code, the expert can rename the test as seen appropriate.

kewisch: You come at the right time, I currently have some eyes on the project and can help get things merged. I think Martin's comment was more about the name of the test, maybe just writing it out, e.g. `stringify property value containing "escaped" semicolons, commas, colons`.

dilyanpalauzov: Once this is going to be merged, please change the description as seems good.

lib/ical/stringify.js
```diff
@@ -209,16 +208,16 @@ stringify.property = function(property, designSet, noFold) {
...
  * in double quotes.
  *
- * @function ICAL.stringify.paramPropertyValue
+ * @function ICAL.stringify.propertyParameterValue
  * @param {String} value      Raw property value
  * @param {boolean} force     If value should be escaped even when unnecessary
  * @return {String}           Given or escaped value when needed
  */
-stringify.paramPropertyValue = function(value, force) {
+stringify.propertyParameterValue = function(value, force) {
```
kewisch: This would break the API and we'd have to bump the major version to be true to semver. I'd suggest we avoid that for now.

kewisch: Noted this for later in #663

kewisch (approved):

PR #659, opened 2024-04-10
Make ical.es5.cjs an UMD module for browsers

### LINKED ISSUE #653: Make browser usage more clear/compatible
There are some breaking changes w.r.t. imports and browsers in 2.0.0. You can't simply <script src> import it there, you'll need
```html
<script type="module">
import ICAL from "https://unpkg.com/ical.js";

// ...
</script>
```

It isn't currently possible to simply global import `ICAL`, even with the es5 version. Tasks here:

1) Update README with a section on browser use
2) Update es5 version to work like before where there is an ICAL global object.

kewisch:
In version 2.0.0 ical.es5.cjs and ical.es5.min.cjs was a cjs module, which made it work with older version of node and use of `require()`. The regular version uses ES6 modules, so it needs to be imported instead of using a script tag. Therefore, none of the versions could be included in a <script> tag in the browser. You can now use the es5 version if you require <script> tag inclusion and would like an `ICAL` global on the window, and also if you require a cjs module in your node project.

[fixes #653]

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,22 +23,47 @@ and example links.
 There is also a validator that demonstrates how to use the library in a webpage in the
 [tools/](https://github.com/kewisch/ical.js/tree/main/tools) subdirectory.
 
-[Try the validator online](http://kewisch.github.io/ical.js/validator.html), it always uses the latest release of ICAL.js.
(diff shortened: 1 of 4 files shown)


mschroeder (approved):
LGTM, but conflict has to be resolved.

PR #665, opened 2024-04-15
Fix stack trace on custom ParserError

kewisch:
I caught a ParserError with a weird stack, and it turns out the custom mangling we did before doesn't really cut it. With ES6+ we can simplify the instance variables.

I've only tested this in recent node and the firefox error console, both seem to contain a stack. The node stack even skips over the error class even without Error.captureStackTrace.

Does this look ok?

### DIFF
--- a/lib/ical/parse.js
+++ b/lib/ical/parse.js
@@ -93,22 +93,8 @@ parse.component = function(str) {
  * @extends {Error}
  * @class
  */
-
 class ParserError extends Error {
-  constructor(message) {
-    super(message);
-    this.name = this.constructor.name;
-
-    try {
-      throw new Error();
-    } catch (e) {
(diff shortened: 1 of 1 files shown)


darktrojan (commented):
The JSDoc needs to be updated, but otherwise this seems okay to me.

kewisch:
@darktrojan What would you like changed in the jsdoc? I think it is still accurate. I guess removing the `@param` so it rather inherits from `Error` maybe?

darktrojan:
Yeah, that's what I had in mind. The `@extends` implies the `@param`. Today I am less convinced it matters. I'll approve the change and you can decide whether you want to keep it.

darktrojan (approved):

PR #666, opened 2024-04-16
Make occurrences in chronological order even if BYMONTHDAY or BYYEARDAY is not in order

darktrojan:
The recurrence iterator has a problem where it makes occurrences in the same order as given in BYMONTHDAY or BYYEARDAY, which doesn't make sense - it should always return them in chronological order.

This patch fixes it by pre-ordering the days that can be used when the month or year changes. A day that is valid for one month/year may be different or not valid for the next month/year, so the days need to be recomputed each time.


lib/ical/recur_iterator.js
```diff
@@ -262,77 +262,79 @@ class RecurIterator {
...
-        this.last = initLast.clone();
-        let [pos, dow] = this.ruleDayOfWeek(bydow);
-        let dayOfMonth = this.last.nthWeekDay(dow, pos);
-
-        // If |pos| >= 6, the byday is invalid for a monthly rule.
-        if (pos >= 6 || pos <= -6) {
-          throw new Error("Malformed values in BYDAY part");
-        }
+    if (this.rule.freq == "MONTHLY") {
+      if (this.has_by_data("BYDAY")) {
```
darktrojan, line 266: Github's UI has failed to figure out that all I've done to this block is indent it. I'm pretty sure the block below (`else if (this.has_by_data("BYMONTHDAY")`) should also only be used if the rule is monthly.

lib/ical/recur_iterator.js
```diff
@@ -262,77 +262,79 @@ class RecurIterator {
...
+          this.last.day = normalized[0];
+          this.by_data.BYMONTHDAY = normalized;
+        } else {
+          // There's no occurrence in this month, find the next valid month.
+          // The longest possible sequence of skipped months is February-April-June,
+          // so we might need to call next_month up to three times.
+          if (!this.next_month() && !this.next_month() && !this.next_month()) {
+            throw new Error("No possible occurrences");
+          }
         }
```
darktrojan, line 337: I've thrown out an earlier hack and just used the valid days in the starting month if there is at least one of them.

lib/ical/recur_iterator.js
```diff
@@ -513,7 +515,10 @@ class RecurIterator {
     let rule;
 
     for (; ruleIdx < len; ruleIdx++) {
-      rule = rules[ruleIdx];
+      rule = parseInt(rules[ruleIdx], 10);
+      if (isNaN(rule)) {
+        throw new Error('Invalid BYMONTHDAY value');
+      }
```
darktrojan, line 521: I found that if the rule is a string, weird things happen.

lib/ical/recur_iterator.js
```diff
@@ -844,6 +850,13 @@ class RecurIterator {
       this.days_index = 0;
       do {
         this.increment_year(this.rule.interval);
+        if (this.has_by_data("BYMONTHDAY")) {
+          this.by_data.BYMONTHDAY = this.normalizeByMonthDayRules(
+            this.last.year,
+            this.last.month,
+            this.rule.parts.BYMONTHDAY
+          );
+        }
```
darktrojan, line 859: This is the important bit (and the same code just below) – go back to the rule as specified and recalculate the valid days for the new month. This deals with negative rules and puts them all in order.

lib/ical/recur_iterator.js
```diff
@@ -1177,6 +1200,14 @@ class RecurIterator {
     } else {
       this.days = [];
     }
+
+    let daysInYear = Time.isLeapYear(aYear) ? 366 : 365;
+    this.days.sort((a, b) => {
+      if (a < 0) a += daysInYear + 1;
+      if (b < 0) b += daysInYear + 1;
+      return a - b;
+    });
```
darktrojan, line 1209: And this does the same for BYYEARDAY rules.

kewisch (approved):
lgtm

PR #670, opened 2024-04-21
recur_expansion: EXDATE can be DATE and DTSTART can be DATE-TIME

dilyanpalauzov:
When EXDATE is in DATE format and DTSTART is DATE-TIME, then to determine whether an occurrence shall be excluded, the occurrence shall be converted to DATE and then compared to EXDATE.

This applies also for the very first EXDATE, when it coincides with DTSTART.

I had this change locally since 2022.  I have not uploaded it, as I have lost hope that PRs here will be handled.


lib/ical/recur_expansion.js
```diff
@@ -198,6 +198,21 @@ class RecurExpansion {
...
+   * and the first parameter is date-with-time, strip the time and compare only
+   * the days.
+   *
+   * @private
+   * @param {ICAL.Time} a   The one object to compare
+   * @param {ICAL.Time} b   The other object to compare
+   */
+  _compare_special(a, b) {
+    if (!a.isDate && b.isDate)
+      return new Time({ year: a.year, month: a.month, day: a.day }).compare(b);
```
kewisch, line 211: This also strips out the timezone, which might affect the comparison. Can you add zone handling here?

kewisch: A test or two around this case and zone handling would also be useful.

kewisch (changes requested):
Thanks for the patch! I'm going to need a few tests for this one, would appreciate if you could add them.

kewisch:
@dilyanpalauzov Did you have a chance to look into this? Would appreciate if you could take care of the review comments.

lib/ical/recur_expansion.js
dilyanpalauzov: I have added a test.

I am not going to add time zone handling, as  I do not understand how it would work.  This patch fixes a problem I faced in 2022.  I do not work with iCalendar in JavaScript currently.

dilyanpalauzov: > This also strips out the timezone, which might affect the comparison. Can you add zone handling here?

The commit message contains:

>  When EXDATE is in DATE format and DTSTART is DATE-TIME, then  to determine whether an occurrence shall be excluded, the occurrence shall be converted to DATE and then compared to EXDATE.

I would say in the conversion of VALUE=DATE-TIME to VALUE=DATE the timezone is irrelevant.

dilyanpalauzov (commented):
I do not know how these reviews of my own change works. Apparently I have to add some comment for the sake of having any comment

dilyanpalauzov:
?!?!

kewisch (approved):

Issue #707, opened 2024-06-15
Invalid typescript declarations


dilyanpalauzov:
I tried the new type declarations from ical.js, originating from https://github.com/kewisch/ical.js/pull/662:

I do `import ICAL from "ical.js"`.  The errors I get are:

In dist/component.d.ts:65: `readonly get name(): string;` → ../ical.js/dist/types/component.d.ts(65,5): error TS1024: 'readonly' modifier can only appear on a property declaration or index signature.

In event.d.ts:42: `constructor(component?: Component | undefined, options: {strictExceptions: boolean; exceptions: Array<Component | Event>; });`→ error TS1016: A required parameter cannot follow an optional parameter.  Indeed, component is optional, options is not.

period.d.ts:41:`static fromJSON(aData: Array<string, string>, aProp: Property, aLenient: boolean): Period;` → error TS2314: Generic type 'Array<T>' requires 1 type argument(s).

design.d.ts:10: `export function getDesignSet(componentName: string): ICAL.design.designSet;` →error TS2503: Cannot find namespace 'ICAL'.

Full list of errors:

../ical.js/dist/types/component.d.ts(65,5): error TS1024: 'readonly' modifier can only appear on a property declaration or index signature.
../ical.js/dist/types/design.d.ts(10,58): error TS2503: Cannot find namespace 'ICAL'.
../ical.js/dist/types/design.d.ts(16,22): error TS2503: Cannot find namespace 'ICAL'.
../ical.js/dist/types/design.d.ts(21,23): error TS2503: Cannot find namespace 'ICAL'.
../ical.js/dist/types/design.d.ts(26,24): error TS2503: Cannot find namespace 'ICAL'.
../ical.js/dist/types/event.d.ts(42,52): error TS1016: A required parameter cannot follow an optional parameter.
../ical.js/dist/types/event.d.ts(174,5): error TS1024: 'readonly' modifier can only appear on a property declaration or index signature.
../ical.js/dist/types/parse.d.ts(26,55): error TS2503: Cannot find namespace 'ICAL'.
../ical.js/dist/types/parse.d.ts(68,59): error TS2304: Cannot find name 'Numeric'.
../ical.js/dist/types/parse.d.ts(93,129): error TS2503: Cannot find namespace 'ICAL'.
../ical.js/dist/types/period.d.ts(41,28): error TS2314: Generic type 'Array<T>' requires 1 type argument(s).
[...]

PR #712, opened 2024-06-17
Fix remaining typescript issues

jannikac:
Hi,
according to #707 there are some remaining issues with the generated typescript types. This PR aims to fix everything except those regarding `ICAL.design.designSet` since that is currently broken. Some commits need an explanation:

- Remove `@readonly` tags because it causes this error in typescript: `readonly' modifier can only appear on a property declaration or index signature.`. When the getters get converted to typescript, typescript changes `@readonly` to the typescript keyword `readonly` which is only allowed in index signatures or properties. However you use the `@readonly` tag on getter functions which is not valid in typescript. I couldn't find a way to omit specific valid jsdoc comments from being parsed by typescript (`@ts-ignore` doesnt work and `@ignore` removed the function from the docs).
- Made the options object optional because otherwise it causes `A required parameter cannot follow an optional parameter.`. Also I took a look at the code and believe that `options` itself and their properties should be optional anyways since `options.strictExceptions` is set in the property declaration to false and  `options.exceptions` states that "If not specified exceptions will automatically be set in relation of compoent's parent"

### DIFF
--- a/lib/ical/timezone_service.js
+++ b/lib/ical/timezone_service.js
@@ -4,6 +4,8 @@
  * Portions Copyright (C) Philipp Kewisch */
 
 import Timezone from "./timezone.js";
+// needed for typescript type resolution
+// eslint-disable-next-line no-unused-vars
 import Component from "./component.js";
 
 let zones = null;
--- a/lib/ical/timezone_service.js
+++ b/lib/ical/timezone_service.js
@@ -99,7 +99,7 @@ const TimezoneService = {
     }
 
     if (!timezone.tzid && !name) {
-      throw new TypeError("neither a timezone nor a name was passed");
+      throw new TypeError("Neither a timezone nor a name was passed");
     }
 
     if (timezone instanceof Timezone) {
--- a/lib/ical/timezone_service.js
+++ b/lib/ical/timezone_service.js
@@ -98,7 +98,7 @@ const TimezoneService = {
       }
     }
 
-    if (!timezone.tzid && !name) {
+    if (!name) {
       throw new TypeError("Neither a timezone nor a name was passed");
     }
 
--- a/lib/ical/timezone_service.js
+++ b/lib/ical/timezone_service.js
@@ -69,22 +69,37 @@ const TimezoneService = {
   /**
    * Registers a timezone object or component.
    *
+   * @param {Component|Timezone} timezone
+   *        The initialized zone or vtimezone.
+   *
    * @param {String=} name
    *        The name of the timezone. Defaults to the component's TZID if not
    *        passed.
-   * @param {Component|Timezone} timezone
-   *        The initialized zone or vtimezone.
    */
-  register: function(name, timezone) {
+  register: function(timezone, name) {
     if (zones === null) {
       this.reset();
     }
 
-    if (name instanceof Component) {
-      if (name.name === 'vtimezone') {
-        timezone = new Timezone(name);
+    // This avoids a breaking change by the change of argument order
+    // TODO remove in v3
+    if (typeof timezone === "string" && name instanceof Timezone) {
+      [timezone, name] = [name, timezone];
+    }
+
+    if (!name) {
+      if (timezone instanceof Timezone) {
         name = timezone.tzid;
+      } else {
+        if (timezone.name === 'vtimezone') {
+          timezone = new Timezone(timezone);
+          name = timezone.tzid;
+        }
       }
+    }
+
+    if (!timezone.tzid && !name) {
+      throw new TypeError("neither a timezone nor a name was passed");
     }
 
     if (timezone instanceof Timezone) {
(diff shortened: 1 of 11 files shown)


jannikac:
Managed to fix all the remaining typescript errors, even the `designSet` ones!

PS: I have updated the tests, so everything passes, although I'm not too familiar with them, please check, if I did that correctly. Also I could open a separate PR for the breaking change if you want, however then one typescript issue will remain unfixed after merging this PR.

lib/ical/timezone_service.js
```diff
@@ -69,25 +78,23 @@ const TimezoneService = {
...
-   * @param {String=} name
-   *        The name of the timezone. Defaults to the component's TZID if not
-   *        passed.
-   * @param {Component|Timezone} timezone
-   *        The initialized zone or vtimezone.
+   * @param {NamedTimezoneOrComponent} namedTimezoneOrComponent
+   *        The name of the timezone and the timezone or a component
    */
-  register: function(name, timezone) {
+  register: function(namedTimezoneOrComponent) {
```
kewisch: I'm not a big fan of objects for arguments unless it is very many, as in many cases consumers need to create an object solely for the purpose of passing arguments and then it is destroyed again.

I think we should make it as follows:
* First argument is `Component|Timezone` and required.
* Second argument is String (name) and optional
* If no name is specified, then name is assumed to be timezone.tzid, regardless of if a timezone is passed or a component that is turned into a timezone.
* Runtime error thrown if neither timezone.tzid is set nor name is passed.
* To avoid a breaking change over argument order, a runtime check to see if arg1 is string and arg2 is Timezone, then just swap them correctly. Add a note to the v3 issue to remove the hack when we are ready for v3. Technically it already breaks typescript consumers, but we didn't have types a few weeks ago and the type checks would catch it.

kewisch (changes requested):

dilyanpalauzov:
I tried the newest types.

My code so far used `new ICAL.Duration({days: -1})` without a problem.  But the types say this is now invalid, as it misses `isNegative`, `weeks`, `hours`, … elements.  Are `hours`, `weeks` required in the object, passed to the constructor?

I guess the parameter of `event.iterator()` should be optional.

In `Time.fromJSDate()` the second parameter used to be optional - skipping it, was equivalent to passing “false”.  Now this parameter is mandatory. Can it be made optional/default value to be `false`?

While the types enforce that Component.getFirstPropertyValue() returns string, in my existing code I assume that getFirstPropertyValue('dtend' / 'duration' / 'rrule' ) return Time / Duration / Recur object.  Please check, if the return type can indeed be non-string?

Isn’t `getAllSubcomponents(name?: string | undefined): Component[];` redundant - name is either optional, or is (a string or undefined)?

CeeMaxx000 (approved):

CeeMaxx000 (approved):

CeeMaxx000 (approved):

CeeMaxx000 (approved):

jannikac:
Hi. I just updated the register function with your suggestion. Should I add a test for the new runtime error aswell?

lib/ical/timezone_service.js
```diff
@@ -69,24 +69,39 @@ const TimezoneService = {
...
+      } else {
+        if (timezone.name === 'vtimezone') {
+          timezone = new Timezone(timezone);
+          name = timezone.tzid;
+        }
       }
     }
 
+    if (!timezone.tzid && !name) {
+      throw new TypeError("neither a timezone nor a name was passed");
```
kewisch:
```suggestion
      throw new TypeError("Neither a timezone nor a name was passed");
```

lib/ical/timezone_service.js
```diff
@@ -69,24 +69,39 @@ const TimezoneService = {
...
         name = timezone.tzid;
+      } else {
+        if (timezone.name === 'vtimezone') {
+          timezone = new Timezone(timezone);
+          name = timezone.tzid;
+        }
       }
     }
 
+    if (!timezone.tzid && !name) {
```
kewisch:
```suggestion
    if (!name) {
```
You already check for the timezone id in the block above and we require the name below.

kewisch (approved):
Fixing the build warning and these two minor comments and we're good, thank you!

kewisch:
> Hi. I just updated the register function with your suggestion. Should I add a test for the new runtime error aswell?

Tests are always well received and maintain coverage, so this would be great!

jannikac:
Actually I think this test case already covers it, so I didn't add another one:
```js
test('with only invalid component', function() {
      assert.throws(function() {
        let comp = new ICAL.Component('vtoaster');
        subject.register(comp);
      }, "neither a timezone nor a name was passed");
    });
```

Accepted your changes and fixed the warning with an unused import, so now we should be good

kewisch:
Thank you so much, I've merged the changes and will make a note in the v3 issue.

Issue #723, opened 2024-06-29
Invalid typescript declarations


dilyanpalauzov:
As mentioned at https://github.com/kewisch/ical.js/pull/712#issuecomment-2179337109 I tried the newest types.

My code so far used `new ICAL.Duration({days: -1})` without a problem.  But the types say this is now invalid, as it misses `isNegative`, `weeks`, `hours`, … elements.  Are `hours`, `weeks` required in the object, passed to the constructor?

I guess the parameter of `event.iterator()` should be optional.

In `Time.fromJSDate()` the second parameter used to be optional - skipping it, was equivalent to passing “false”.  Now this parameter is mandatory. Can it be made optional/default value to be `false`?

While the types enforce that Component.getFirstPropertyValue() returns string, in my existing code I assume that getFirstPropertyValue('dtend' / 'duration' / 'rrule' ) return Time / Duration / Recur object.  Please check, if the return type can indeed be non-string?

Isn’t `getAllSubcomponents(name?: string | undefined): Component[];` redundant - name is either optional, or is (a string or undefined)?

jannikac:
Hi,

> My code so far used new ICAL.Duration({days: -1}) without a problem. But the types say this is now invalid, as it misses isNegative, weeks, hours, … elements. Are hours, weeks required in the object, passed to the constructor?

> I guess the parameter of event.iterator() should be optional.

> In Time.fromJSDate() the second parameter used to be optional - skipping it, was equivalent to passing “false”. Now this parameter is mandatory. Can it be made optional/default value to be false?

> While the types enforce that Component.getFirstPropertyValue() returns string, in my existing code I assume that getFirstPropertyValue('dtend' / 'duration' / 'rrule' ) return Time / Duration / Recur object. Please check, if the return type can indeed be non-string?

I believe the jsdoc are indeed incorrect. I created a PR to fix that.

> Isn’t getAllSubcomponents(name?: string | undefined): Component[]; redundant - name is either optional, or is (a string or undefined)?

This is a bit redundant because the `?` means the type `string?` is defined to be of `string | undefined`. We currently document the every optional parameter with this syntax `@param {=Number} optionalParameter` which causes it to be documented in typescript with `optionalParameter?: number | undefined`. Documenting this with `@param {Number} [optionalParameter]` creates `optionalParameter?: number` in typescript which would be clearer. However we'd have to change a lot of code to make this a consistent change, maybe @kewisch can decide?

dilyanpalauzov:
[...]

PR #727, opened 2024-07-01
Fix incorrect jsdocs

jannikac:
Fixes for #723.


lib/ical/component.js
```diff
@@ -278,7 +278,7 @@ class Component {
    * Returns first property's value, if available.
    *
    * @param {String=} name    Lowercase property name
-   * @return {?String}        The found property value.
+   * @return {?Object}        The found property value.
```
dilyanpalauzov: This is an improvement, but if possible, the types should be spelled in a more narrow way.

jannikac: Hm but as far as I understand it the hydrateValue function can return any wrapped Object or null. But im not sure which objects it can return exactly. I read into the spec a bit and all values that match classes are the following:
- Binary
- Duration
- Period
- Recur
- Time

So these can be returned from getFirstValue as a union correct?

dilyanpalauzov: I tryed to add the above types myself, to see if it will work as expected.  However, when I insert into component.js:
```javascript
 * @typedef {import("./duration.js).Duration} Duration
 * @typedef {import("./time.js).Time} Time
 * @typedef {import("./recur.js).Recur} Recur
 ```
I get the error

> (!) [plugin typescript] lib/ical/component.js (24:7): @rollup/plugin-typescript TS2451: Cannot redeclare block-scoped variable 'NAME_INDEX'.

Comment (dilyanpalauzov) [reply lib/ical/component.js]: The possible return types are defined in RFC 5545 in the Table of Content under [3.3](https://datatracker.ietf.org/doc/html/rfc5545#section-3.3)  “Property Value Data Types”.

Comment (dilyanpalauzov) [reply lib/ical/component.js]: Oh, I messed the brackets above.  With correct brackets there is no such warning.

Comment (dilyanpalauzov) [reply lib/ical/component.js]: I tried, but I do not know how to get these types right, unless I spend much more time.  I suggest you include the types. Then I will try the change out and let you know if it works.

Comment (jannikac) [reply lib/ical/component.js]: Just pushed what I believe to be the correct types. You can use those by following the steps in this comment: https://github.com/kewisch/ical.js/issues/719#issuecomment-2196668917

Comment (dilyanpalauzov) [reply lib/ical/component.js]: I think `string` is also valid value for `getFirstPropertyValue("SUMMARY")` / `getFirstValue()`.

Comment (dilyanpalauzov) [reply lib/ical/component.js]: The value of the GEO property is two semicolon-separated float values.  How is this mapped to types for `getFirstPropertyValue('geo')`?

Comment (jannikac) [reply lib/ical/component.js]: I added string to the return types.

I don't think a type geo exists in the current implementation. However, I could add a typedef for that if its needed.

Comment (dilyanpalauzov) [reply lib/ical/component.js]: Please add some type for the GEO property value.  I use it and currenlty it filers a lot of alarms, as it does not match type (nothing from that union)

Comment (dilyanpalauzov) [reply lib/ical/component.js]: With the most recent change, when I do
```javascript
import ICAL from 'ical.js'
…
let geo = subcomponent.getFirstPropertyValue('geo') as ICAL.Geo
```
the error I get is:

> error TS2694: Namespace '_default' has no exported member 'Geo'.

Indeed, dist/types/module.ts.d has
```
declare namespace _default {
...
}
export default _default;
```
and Geo is not mentioned in the namespace.

dilyanpalauzov: I think ICAL.Recur must have properties `bymonth`, `bymonthday`, `byday`, etc.  My code relayed on this so far.

stj-be: > With the most recent change, when I do
> 
> ```js
> import ICAL from 'ical.js'
> …
> let geo = subcomponent.getFirstPropertyValue('geo') as ICAL.Geo
> ```
> 
> the error I get is:
> 
> > error TS2694: Namespace '_default' has no exported member 'Geo'.
> 
> Indeed, dist/types/module.ts.d has
> 
> ```
> declare namespace _default {
> ...
> }
> export default _default;
> ```
> 
> and Geo is not mentioned in the namespace.

So the issue here is that `@memberof` is not supported in typescript so something like `ICAL.customTypedef` is not possible AFAIK. To use the geo type you need to pull the latest PR and import it like this `import { Geo } from "{IcalPAckage}/dist/types/types";`

> I think ICAL.Recur must have properties bymonth, bymonthday, byday, etc. My code relayed on this so far.

What exactly are you referring to here? The constructor has those properties in the data array.
`let recur = new ICAL.Recur({ bymonthday: [20] });`
Or are you referring to the Class instance which should have those properties?

jannikac: > With the most recent change, when I do
> 
> ```js
> import ICAL from 'ical.js'
> …
> let geo = subcomponent.getFirstPropertyValue('geo') as ICAL.Geo
> ```
> 
> the error I get is:
> 
> > error TS2694: Namespace '_default' has no exported member 'Geo'.
> 
> Indeed, dist/types/module.ts.d has
> 
> ```
> declare namespace _default {
> ...
> }
> export default _default;
> ```
> 
> and Geo is not mentioned in the namespace.

So the issue here is that `@memberof` is not supported in typescript so something like `ICAL.customTypedef` is not possible AFAIK. To use the geo type you need to pull the latest PR and import it like this `import { Geo } from "{IcalPAckage}/dist/types/types";`

> I think ICAL.Recur must have properties bymonth, bymonthday, byday, etc. My code relayed on this so far.

What exactly are you referring to here? The constructor has those properties in the data array.
`let recur = new ICAL.Recur({ bymonthday: [20] });`
Or are you referring to the Class instance which should have those properties?

dilyanpalauzov:
```javascript
import type { Geo } from 'ical.js/dist/types/types'
```
is suboptimal, but does work.  

> Or are you referring to the Class instance which should have those properties?

Yes the class instance can have these properties, this is currently not reflected by the type system.

jannikac: I assume you refer to the `recur.parts` property, correct? I couldn't find the properties you mentioned in any other place. If my assumption is correct, I the type for that should work.

dilyanpalauzov: > I think ICAL.Recur must have properties bymonth, bymonthday, byday, etc. My code relayed on this so far.

The above statement is not correct.  What it meant was, that for x of type ICAL.Recur, x.bymonth can exist.

I have used ical.js in my code for many use cases without type validation.  What my code did was to convert a variable of type ICAL.Recur to JSON, by calling `rrule = rrule.toJSON()`.  Afterwards `rrule.bymonth` was a valid query.   After I switched to the types from ical.js that expression failed.

Irrespective of this I think the recent changes on type Recur are nice.

jannikac: Yep, agree aswell. Did you find any more typescript errors we can fix in this PR perhaps?

dilyanpalauzov: > Did you find any more typescript errors we can fix in this PR perhaps?

No, but I have also not searched for more.  I have to add explicit types on many places.  I suggest the changes here go upstream.  If I find more nits, I will file a new ticket.

kewisch (approved):

PR #842, opened 2025-03-11
doc fix

alexler12345:
just fixed the two design entry bug on the gh pages


kewisch:
The fix needs to be in the jsdoc in the source code, the gh pages are generated. If you just change it here then it will be overwritten on the next docs deploy. Could you check if there is a way to change this in the source code instead?

kewisch:
Closing this for now, once you've had a chance please feel free to resubmit!

PR #856, opened 2025-04-23
iCalendar Duration RFC Compliance Fix

AksPu:
Previously, duration was not compliant with RFC Guidelines, leading to synchronization errors. We have modified the toString function in duration.js to ensure RFC compliant duration, and prevent synchronization errors.

Issue Link: https://github.com/kewisch/ical.js/issues/732

### DIFF
--- a/lib/ical/duration.js
+++ b/lib/ical/duration.js
@@ -273,15 +273,35 @@ class Duration {
       let str = "";
       if (this.isNegative) str += "-";
       str += "P";
-      if (this.weeks) str += this.weeks + "W";
-      if (this.days) str += this.days + "D";
-
-      if (this.hours || this.minutes || this.seconds) {
-        str += "T";
-        if (this.hours) str += this.hours + "H";
-        if (this.minutes) str += this.minutes + "M";
-        if (this.seconds) str += this.seconds + "S";
+      let hasWeeks = false;
+      if (this.weeks) {
+        if (this.days || this.hours || this.minutes || this.seconds) {
+          str += (this.weeks * 7 + this.days) + "D";
+        } else {
+          str += (this.weeks + "W");
+          hasWeeks = true;
+        }
+      } else if (this.days) {
+        str += (this.days + "D");
       }
+
+      if (!hasWeeks) {
+        if (this.hours || this.minutes || this.seconds) {
+          str += "T";
+          if (this.hours) {
+            str += this.hours + "H";
+          }
+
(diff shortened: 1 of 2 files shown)


rrishi-r:
If someone could review this PR / our modification when they get a chance, that would be great. Thank you!

lib/ical/duration.js
```diff
@@ -273,8 +273,20 @@ class Duration {
       let str = "";
       if (this.isNegative) str += "-";
       str += "P";
-      if (this.weeks) str += this.weeks + "W";
-      if (this.days) str += this.days + "D";
+      console.log(this.weeks)
+      if(this.weeks){
+        if (this.days || this.hours || this.minutes || this.seconds){
+          str += (this.weeks*7 + this.days + "D");
```
kewisch: Can you run the linter on your code? I see a few things I think it might complain on re spacing and brackets.

I think we should go with this:

```javascript
str += (this.weeks * 7 + this.days) + "D";
```

to make a the arithmetic addition a bit more clear

lib/ical/duration.js
```diff
@@ -273,8 +273,20 @@ class Duration {
       let str = "";
       if (this.isNegative) str += "-";
       str += "P";
-      if (this.weeks) str += this.weeks + "W";
-      if (this.days) str += this.days + "D";
+      console.log(this.weeks)
```
kewisch: Please remove the console log statement

lib/ical/duration.js
```diff
@@ -273,8 +273,20 @@ class Duration {
...
+          str += (this.weeks + "W");
+        }
+      }
+      else{
+        if(this.days){
+          str += (this.days + "D");
+        }
+      }
 
       if (this.hours || this.minutes || this.seconds) {
```
kewisch: Looking at the linked issue it sounds like dur-week is mutually exclusive with dur-time. If we want to produce valid values we should probably also take care of this.

kewisch (changes requested):
Could you add a test in addition to these comments?

rrishi-r:
Thanks for the feedback - we will work on making these changes later today!

AksPu:
Thanks for the feedback. I have made another commit to address your review. Please let me know if it looks okay!

kewisch:
There are still various linter errors, can you use `npm run lint` to check?

kewisch (approved):

PR #863, opened 2025-05-15
Fix support for parsing UTC Dates with trailing decimal seconds

Patronics:
This allows date-time strings to be specified in the native way that native JS `Date.toISOString()` and `Date.toJSON()` are formatted (in the form `2025-05-15T03:09:02.000Z`, for example). They were already correctly imported, except that the trailing Z to specify UTC wasn't detected, since it was hardcoded to check only the 19th character, so they were interpreted as floating time.

This fix also means the format used in the newer Temporal API's `instant.toJSON()` and `instant.toString()` are supported as well.

In case there's any other software that appends something else to the string and depends on the Z being at the 19th character, it still handles that case as well, though that can of course be easily adjusted if that behavior is not desired.

### DIFF
--- a/lib/ical/time.js
+++ b/lib/ical/time.js
@@ -184,7 +184,7 @@ class Time {
     let zone;
     let zoneId;
 
-    if (aValue[19] && aValue[19] === 'Z' || aValue.slice(-1) === 'Z') {
+    if (aValue.slice(-1) === 'Z') {
       zone = Timezone.utcTimezone;
     } else if (prop) {
       zoneId = prop.getParameter('tzid');
--- a/lib/ical/time.js
+++ b/lib/ical/time.js
@@ -184,7 +184,7 @@ class Time {
     let zone;
     let zoneId;
 
-    if (aValue[19] && aValue[19] === 'Z') {
+    if (aValue[19] && aValue[19] === 'Z' || aValue.slice(-1) === 'Z') {
       zone = Timezone.utcTimezone;
     } else if (prop) {
       zoneId = prop.getParameter('tzid');
--- a/test/time_test.js
+++ b/test/time_test.js
@@ -255,6 +255,33 @@ suite('icaltime', function() {
     });
   });
 
+  suite('#fromDateTimeString', function() {
+    test('utc without decimals', function() {
+      let date = "2012-01-01T00:00:00Z";
+      let expected = date;
+      let subject = Time.fromDateTimeString(date);
+      assert.equal(
+        subject.toString(),
+        expected);
+    });
+    test('utc with decimals', function() {
+      let date = "2012-01-01T00:00:00.000Z";
+      let expected = new Date(date);
+      let subject = Time.fromDateTimeString(date);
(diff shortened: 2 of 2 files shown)


lib/ical/time.js
```diff
@@ -180,7 +180,7 @@ class Time {
     let zone;
     let zoneId;
 
-    if (aValue[19] && aValue[19] === 'Z') {
+    if (aValue[19] && aValue[19] === 'Z' || aValue.slice(-1) === 'Z') {
```
kewisch: The hardcoding was due to performance optimizations back then. I think we can simplify this to `if (aValue.endsWith('Z'))`. Could you also add some tests to cover the different cases that could occur here?

kewisch (changes requested):

lib/ical/time.js
Patronics: based on [this testing](https://www.measurethat.net/Benchmarks/Show/34556/0/javascript-stringendswith-vs-stringslice-and-triple-equ) the .slice variant is significantly faster than .endsWith and matches the performance of hardcoded indexes, so probably worth keeping. I'll remove the redundant check for index 19 and get you a few tests written for it :)

kewisch: I didn't expect that, interesting results. Does that reproduce consistently for you? I ran it and endsWith had about as much ops/sec as a hardcoded index. slice had substantially less ops/sec and was about 6x slower. https://www.measurethat.net/Benchmarks/ShowResult/605389

Patronics: That seems to reproduce pretty consistently in my testing, but it does vary by browser, now that I'm testing it in more browsers. Here's the results of my testing:
Safari (which I initially tested, both mobile and desktop give similar results): 
    endsWith: Around 80Mops/s
    slice: Around 520Mops/s
    hardcoded: Around 510Mops/s
Chromium:
    endsWith: Around 155Mops/s
    slice: Around 130Mops/s
    hardcoded: Around 170Mops/s
Firefox:
    endsWith: Around 1,570Mops/s
    slice: Around 265Mops/s
    hardcoded: Around 1,570Mops/s
    (occasionally when testing in Firefox endsWith and hardcoded reported "infinity" ops/s, whatever that means?) Seems they've optimized very well for the endsWith and hardcoded case in Firefox and/or the test isn't running properly, perhaps it's caching the result to avoid repeating the same operation?

Given the results you reported, I'm guessing you were testing on Firefox?
My thinking is that slice seems to be the most consistently performant regardless of browser overall, especially given the relative market shares of the three browser families. Slightly slower in Chromium, but not majorly so, and consistently maintaining at least 130,000,000 ops/second on my system regardless of browser, shouldn't be a substantial bottleneck or anything anyway. But I can definitely switch it to endsWith if you prefer, either way works.

Patronics:
Is there anything else you need from me before this can be merged?

kewisch:
I'll take a look very soon, apologies for the delay.

kewisch (approved):

Issue #893, opened 2025-06-30
Regression with 2.2.0 on VCARD: backslash inserted between `;` in `N:`


dilyanpalauzov:
I use ical.js to generate VCARD 3.0.  With version 2.1.0 the output is `N:;A-B;;;`.  After upgrading to version 2.2.0 the output is `N:\;A-B\;\;\;`.  The backslashes are too much.  The code using ical.js is the same in both cases.

dilyanpalauzov:
Property values are added by
```js
let vcard = new ICAL.Component('vcard')
vcard.addPropertyWithValue('adr', `;;;;;;${body.country}`)
vcard.addPropertyWithValue('n', `;${body.name};;;`)
```

kewisch:
Thanks for reporting! Can you confirm it uses the design set for vcard3? #680 is the only issue I see in the list that seems related.

dilyanpalauzov:
I have
```js
let vcard = new ICAL.Component('vcard')
vcard.addPropertyWithValue('version', '3.0')
```
I can neither confirm nor deny what design set is used.  How do I get and set the VCARD 3.0 design set?  I am not parsing input to create a VCard, I am using new Component and addPropertyWithValue().

dilyanpalauzov:
VCARD 3 RFC 2426 and VCARD 4 RFC 6350 use the same semi-colon logic for the N and ADR fields.  For `N:` even the example is the same:
```
        N:Public;John;Quinlan;Mr.;Esq.

        N:Stevenson;John;Philip,Paul;Dr.;Jr.,M.D.,A.C.P.
```
Example for ADR in VCARD 4.0 from RFC 6350:
```

     ADR;GEO="geo:12.3457,78.910";LABEL="Mr. John Q. Public, Esq.\n
      Mail Drop: TNE QB\n123 Main Street\nAny Town, CA  91921-1234\n
      U.S.A.":;;123 Main Street;Any Town;CA;91921-1234;U.S.A.
```
Example for ADR in VCARD 3.0 from RFC 2426:
```
        ADR;TYPE=dom,home,postal,parcel:;;123 Main
          Street;Any Town;CA;91921-1234
```
so the escaple logic for semi-colons is the same.

kewisch:
[...]

PR #904, opened 2025-07-18
Fix #509. Stop iterating months if no occurrences found in 28 years.

### LINKED ISSUE #509: Infinite loop caused by invalid recurrence
Reproducible on latest main.

**To reprocude, we can use the recur_iterator_test.js and add the following code:**

```
testRRULE('FREQ=MONTHLY;BYDAY=1TH;BYSETPOS=2', {
        dtStart: '2015-01-11T08:00:00',
        dates: [
          '2015-02-05T08:00:00'
        ]
      });
```

after that, run

`
grunt test-node:single --test=test/recur_iterator_test.js 
`

PS: I believe the problem in inside the next_month function but I'm  not sure how to fix it.

darktrojan:
In 28 years all possible combinations of month, day of week, and leap year have been tested. If we find no occurrence in that time we're never going to.

The filed bugs give a recurrence rule that is trivially wrong. I could optimise for this, but I've instead gone for a general case which should prevent all infinite loops in monthly recurring rules.

### DIFF
--- a/lib/ical/recur_iterator.js
+++ b/lib/ical/recur_iterator.js
@@ -367,7 +367,7 @@ class RecurIterator {
           // The longest possible sequence of skipped months is February-April-June,
           // so we might need to call next_month up to three times.
           if (!this.next_month() && !this.next_month() && !this.next_month()) {
-            throw new Error("No possible occurrences");
+            throw new InvalidRecurrenceRuleError();
           }
         }
       }
@@ -397,6 +397,7 @@ class RecurIterator {
     }
 
     let valid;
+    let invalid_count = 0;
     do {
       valid = 1;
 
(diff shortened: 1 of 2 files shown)


lib/ical/recur_iterator.js
```diff
@@ -418,6 +419,11 @@ class RecurIterator {
         break;
       case "MONTHLY":
         valid = this.next_month();
+        if (++invalid_count == 336) {
```
kewisch: Would it make sense to add something like this for other rule frequencies as well?

kewisch (approved):
Let me know if you want to add something for the one commented issue

kewisch:
When merged I can cut a new release, I hear you'd like to update ical.js to upstream  When you update keep in mind there is a recent regression with vcard3 that might affect Thunderbird as well. Any help there is appreciated #893.

lib/ical/recur_iterator.js
darktrojan: I'm not sure that it makes sense to. Month is the only unit where we loop like this, which makes sense as months have inconsistent lengths. We might need to loop years for the same reason due to another bug I just found (and because we seem to miss a bunch of `BYSETPOS` edge cases) but for now `next_year` almost always returns 1.

(I've realised I'm doing the wrong thing here if `valid` is truthy, which I'll fix, but it probably doesn't matter.)

darktrojan:
Bah, I thought there was a bug and made a change, but it was right the first time. Hate it when that happens.

PR #906, opened 2025-07-21
Prevent BYYEARDAY=366 from wrapping into the next year.

darktrojan:
The iterator shouldn't return "January 1, year+1" for non-leap years with `BYYEARDAY=366`. Nor should it return "December 31, year-1" for `BYYEARDAY=-366`. I've changed `next_year` to work more like `next_month` and return 0 if there are no matches in the year.

### DIFF
--- a/lib/ical/recur_iterator.js
+++ b/lib/ical/recur_iterator.js
@@ -294,7 +294,12 @@ class RecurIterator {
         throw new InvalidRecurrenceRuleError();
       }
 
-      this._nextByYearDay();
+      // If there's no occurrence in this year, try the following years. This
+      // would only happen looking for day 366 or -366.
+      if (!this._nextByYearDay() && !this.next_year() && !this.next_year() && !this.next_year()) {
+        // This should not be possible, but just in case it is, stop.
+        throw new InvalidRecurrenceRuleError();
+      }
     }
 
     if (this.rule.freq == "MONTHLY") {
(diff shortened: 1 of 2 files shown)


lib/ical/recur_iterator.js
```diff
@@ -428,7 +433,14 @@ class RecurIterator {
         }
         break;
       case "YEARLY":
-        this.next_year();
+        valid = this.next_year();
+        if (valid) {
+          invalid_count = 0;
+        } else if (++invalid_count == 28) {
+          // We've been through all the possible months and not found a recurrence. Stop.
```
darktrojan: Not sure how to word this. This is technically correct but looks wrong.

darktrojan:
I just tried the recurrence tester (good tool, BTW!) and realised we're _not_ changing the year, only wrapping the date. That's even worse. According to the current code, the next five 366th days are:

- 2026-01-01
- 2027-01-01
- 2028-12-31
- 2029-01-01
- 2030-01-01

lib/ical/recur_iterator.js
kewisch: It is fine. Years though, not months?

kewisch (approved):
Feel free to merge with the comment fixed

lib/ical/recur_iterator.js
darktrojan: I meant what I wrote at the time but I've since thought of a better wording.

darktrojan:
Didn't make a mess of that _at all_. 😅

PR #913, opened 2025-08-06
Fix #893, regression in vCard 3 text values.

### LINKED ISSUE #893: Regression with 2.2.0 on VCARD: backslash inserted between `;` in `N:`

darktrojan:
As suggested in #893.


kewisch (approved):
Thanks for taking care, much appreciated!

PR #962, opened 2026-02-24
Add vCard3 date-or-datetime type for BDAY and REV; fixes #961

### LINKED ISSUE #961: vCard 3 BDAY serialised incorrectly
The [vCard 3 spec says](https://datatracker.ietf.org/doc/html/rfc2426#section-3.1.5) BDAY values should be written as:

- `BDAY:1996-04-15`
- `BDAY:1953-10-15T23:10:00Z`
- `BDAY:1987-09-27T08:30:00-06:00`

… but a round-trip though ical.js changes these to:

- `BDAY;VALUE=DATE:19960415`
- `BDAY:19531015T231000Z`
- `BDAY:19870927T083000-0600`

… which is incorrect.

Related: https://bugzilla.mozilla.org/show_bug.cgi?id=1824370

darktrojan:
I _think_ this fixes it cleanly.

Fixes #961


lib/ical/design.js
```diff
@@ -816,9 +816,24 @@ let vcardProperties = extend(commonProperties, {
 });
 
 let vcard3Values = extend(commonValues, {
+  "date-or-datetime": {
```
kewisch: Can you add a comment here noting this isn't yet a valid jcard type?

kewisch (approved):
The trouble is that `date-or-datetime` isn't a valid jcard type, we simply don't have one for this purpose. I think we could make one up and propose an RFC to extend the spec, specifically for types necessary to take care of vcard3.

The value type shouldn't be valid for vcard4. Maybe we can also add a test to make sure that it isn't accidentally used, if we don't already have something like that.

darktrojan:
Ah, that's a thing I hadn't considered at all. I was just trying to fix it. :-) Perhaps there is a better way to do this?

kewisch:
I think we could do this with the `date-and-or-time`, or evn the individual types,  and just make sure it is serialized correctly for the vcard3 case?

Unless there are other date-times or dates in vcard3 that do not have dashes, a `date-time` or `date` type should always serialize to one with dashes. I believe you could change the date-and-or-time serializers in vcard3Values to output the right thing.

darktrojan:
This is better. It doesn't involve inventing things.

kewisch (approved):
Looks great, thank you!

PR #985, opened 2026-05-28
Implement BYSETPOS support for monthly recurrence rules with BYMONTHDAY

### LINKED ISSUE #960: Bug in BYSETPOS
This is regarding the latest tagged release v2.2.1, which is the same as is currently being used (as of writing this, on 2026-02-08) on your tester tool page [https://kewisch.github.io/ical.js/recur-tester.html](https://kewisch.github.io/ical.js/recur-tester.html).

I cannot get any BYSETPOS values to work.

The use case I ran into this with was in trying to create a recurrence rule which effectively meant "Give me the Nth day of each month, or the last day of that month if there are less than N days in that month."

The correct answer to this (for values of N >= 28 anyway, done for brevity's sake in this example), at least according to how I understand the iCalendar/RFC5545 standard, is:
`RRULE:FREQ=MONTHLY;BYMONTHDAY={28..N};BYSETPOS=-1`
where the `{28..N}` is meant to say "the list of numbers from 28 to N"

So if you wanted the 30 day of the month, or the last day of the month for months that don't have that many days, it would be:
`RRULE:FREQ=MONTHLY;BYMONTHDAY=28,29,30;BYSETPOS=-1`

but I was puzzled by what I saw in the results:

Image

As you can see, yes it did produce the correct result, but it was mixed in with invalid results as well. I thought at first I had my logic in constructing the RRULE string wrong, but I've double and triple-checked; it's right. (I checked with other implementations which validate this).

What I've come to realize is that it appears as though the BYSETPOS portion of the rule is simply ignored completely. Look at those results again and think what would happen if that BYSETPOS simply wasn't there... _then_ the results would be 100% correct.

I verified this by giving it a simpler test (notice I changed the start and end dates to be New Year's at midnight to make it as simple as possible to reason about):

Image

And again, it appears as if the BYSETPOS portion of the RRULE is just completely ignored.

mschroeder:
Fixes #960

### DIFF
--- a/lib/ical/recur_iterator.js
+++ b/lib/ical/recur_iterator.js
@@ -353,26 +353,29 @@ class RecurIterator {
           throw new Error("Malformed values in BYDAY part");
         }
       } else if (this.has_by_data("BYMONTHDAY")) {
-        // Change the day value so that normalisation won't change the month.
-        this.last.day = 1;
-
-        // Get a sorted list of days in the starting month that match the rule.
-        let normalized = this.normalizeByMonthDayRules(
-          this.last.year,
-          this.last.month,
-          this.rule.parts.BYMONTHDAY
-        ).filter(d => d >= this.last.day);
-
-        if (normalized.length) {
-          // There's at least one valid day, use it.
-          this.last.day = normalized[0];
-          this.by_data.BYMONTHDAY = normalized;
+        if (this.has_by_data("BYSETPOS")) {
+          this._initMonthlyByMonthDaySetPos();
         } else {
-          // There's no occurrence in this month, find the next valid month.
-          // The longest possible sequence of skipped months is February-April-June,
-          // so we might need to call next_month up to three times.
-          if (!this.next_month() && !this.next_month() && !this.next_month()) {
-            throw new InvalidRecurrenceRuleError();
+          // Change the day value so that normalisation won't change the month.
+          this.last.day = 1;
+          // Get a sorted list of days in the starting month that match the rule.
+          let normalized = this.normalizeByMonthDayRules(
+            this.last.year,
+            this.last.month,
+            this.rule.parts.BYMONTHDAY
+          ).filter(d => d >= this.last.day);
+
+          if (normalized.length) {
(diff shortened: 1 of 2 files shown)


lib/ical/recur_iterator.js
```diff
@@ -1359,6 +1366,92 @@ class RecurIterator {
...
+   * Advance MONTHLY iteration with BYMONTHDAY and BYSETPOS.
+   *
+   * @private
+   * @return {Number} 1 if a valid day was found, 0 otherwise
+   */
+  _nextMonthlyByMonthDaySetPos() {
+    let day = this._findByMonthDaySetPos(this.last.day + 1);
+    let attempts = 0;
+
+    while (!day && attempts++ < 48) {
```
kewisch, line 1435: Could you add a comment on why 48 attempts, here and in the last function?

lib/ical/recur_iterator.js
```diff
@@ -802,28 +805,32 @@ class RecurIterator {
         }
       }
     } else if (this.has_by_data("BYMONTHDAY")) {
-      this.by_indices.BYMONTHDAY++;
+      if (this.has_by_data("BYSETPOS")) {
+        data_valid = this._nextMonthlyByMonthDaySetPos();
+      } else {
```
kewisch, line 810: Have you considered cases where we have BYMONTHDAY, BYSETPOS, and other rule parts? I think theoretically a BYDAY could appears as well.

kewisch (approved):
Spectacular, thanks for tackling this Martin. Good to see you're still following :)

Some minor comments if you could take a look

hiSandog:
The BYMONTHDAY/BYSETPOS coverage is strong, especially the negative and invalid-position cases. One edge worth considering is interaction with `BYDAY`, because RFC expansion normally filters the candidate set before applying `BYSETPOS`. A test like monthly `BYMONTHDAY=1,2,3,4,5;BYDAY=MO,TU,WE,TH,FR;BYSETPOS=-1` would make sure the selected position is based on the fully filtered monthly candidate list, not only the normalized month days.

PR #1006, opened 2026-06-29
Fix vCard 3 time value to keep ':' separators on serialization

mahirhir:
#962 fixed the vCard 3 `date` and `date-time` value types to serialize with their separators (`-` and `:`), because vCard 3 uses the ISO 8601 extended form (e.g. `REV:1995-10-31T22:27:10Z`). The third member of that group, the standalone `time` value, was left pointing at `icalValues.time`, which strips the colons and assumes the basic `HHMMSS` form.

That breaks both directions for a vCard 3 time:

```js
ICAL.design.vcard3.value.time.fromICAL('15:30:00'); // '15::3:0:'  (mangled)
```

End to end, a property with `VALUE=time` round-trips to a corrupted value:

```
BEGIN:VCARD
VERSION:3.0
FN:Test
X-TIME;VALUE=time:15:30:00
END:VCARD
```

parses and re-serializes as `X-TIME;VALUE=TIME:15:30:` (seconds dropped, colons mangled). Note the time portion of a vCard 3 `date-time` already keeps its colons (`2026-02-24T15:30:00`), so the standalone `time` value was inconsistent with it.

The fix applies the same shape #962 used for `date`/`date-time`: reuse `vcardValues.time` for parsing and decoration, and override `toICAL` to keep the value as-is. With that, `15:30:00`, `15:30:00Z` and `15:30:00-05:00` all round-trip, and the value decorates to a proper `VCardTime`.

Added a `time (vcard3)` suite next to the existing `date (vcard3)` / `date-time (vcard3)` suites. `npm run test` and `npm run lint` pass.

### DIFF
--- a/lib/ical/design.js
+++ b/lib/ical/design.js
@@ -834,7 +834,13 @@ let vcard3Values = extend(commonValues, {
   "phone-number": vcardValues["phone-number"],
   uri: icalValues.uri,
   text: vcardValues.text,
-  time: icalValues.time,
+  time: {
+    // As in vCard, but serialised differently (':' are included).
+    ...vcardValues.time,
+    toICAL: function(aValue) {
+      return aValue;
+    }
+  },
   vcard: icalValues.text,
   "utc-offset": {
     toICAL: function(aValue) {
--- a/test/design_test.js
+++ b/test/design_test.js
@@ -579,6 +579,31 @@ suite('design', function() {
       });
     });
 
+    suite('time (vcard3)', function() {
+      setup(function() {
+        subject = ICAL.design.vcard3.value.time;
+      });
+
+      test('#(to|from)ICAL time', function() {
+        let original = '15:00:00';
+        let fromICAL = subject.fromICAL(original);
+
+        assert.equal(fromICAL, '15:00:00');
+        assert.equal(subject.toICAL(fromICAL), original);
+      });
+
+      test('#(un)decorate time', function() {
+        let undecorated = '15:00:00';
+        let decorated = subject.decorate(undecorated);
+
+        assert.equal(decorated.hour, 15, 'hour');
+        assert.equal(decorated.minute, 0, 'minute');
+        assert.equal(decorated.second, 0, 'second');
+
+        assert.equal(subject.undecorate(decorated), undecorated);
+      });
+    });
+
     suite('duration', function() {
       setup(function() {
         subject = subject.value.duration;


mahirhir:
No review on this since June. Here it is measured, and the measurement says the same thing the source does, so both are below.

**On `main`, this object literal already makes the decision twice.** `vcard3Values` overrides `toICAL` for two of its three date-ish types and not the third:

```js
// lib/ical/design.js:818
let vcard3Values = extend(commonValues, {
  date: {
    // As in vCard, but serialised differently ('-' are included).
    ...vcardValues.date,
    toICAL: function(aValue) { return aValue; }
  },
  "date-time": {
    // As in vCard, but serialised differently ('-' and ':' are included).
    ...vcardValues['date-time'],
    toICAL: function(aValue) { return aValue; }
  },
  ...
  time: icalValues.time,        // <- no override, and not the vCard table either
```

`date` and `date-time` carry a vCard3 override with a comment explaining that vCard 3.0 keeps the separators. `time` points at the iCalendar table, which strips them. This PR adds the same three lines and the same comment for `time`, so it is the existing decision applied to the sibling rather than a new idea.

**What it does to a value**, running the published `ical.js@2.2.1` (`sha256 02b0ab420c2b3af5`):

```
property             in          out(after parse -> toString)
X-NOTE;VALUE=text    hello       hello          preserved
X-TIME;VALUE=time    15:00:00    15:00:         LOST
X-TIME;VALUE=time    15:00:00Z   15:00:         LOST
```

The two functions applied in order explain it:

```
fromICAL("15:00:00") = "15::0:0:"
toICAL("15::0:0:")   = "15:00:"
```

`icalValues.time.fromICAL` expects the compact `HHMMSS` form and inserts colons at fixed offsets. Given a value that already has colons, it inserts more, and `toICAL` then slices at fixed offsets from the mangled string. The result is not a truncation in any meaningful sense; it is two format assumptions applied to each other. The seconds disappear and the output is not a valid time.

**One thing I found while checking, which is yours to decide rather than mine.** My control asserted that `date` round-trips, and it failed:

```
X-WHEN;VALUE=date    2026-09-14  20260914       LOST
```

That is not a second bug. On the published 2.2.1 the `date` and `date-time` overrides are absent entirely, so all three types reformat. Those two are already fixed on `main` and have not been released; `time` is the one still open there. So a reader on 2.2.1 sees three broken types, a reader of `main` sees one, and this PR closes the last one.

I nearly reported the `date` row as a separate finding before checking which side of the release boundary it sat on.

One thing I am not claiming:

```console
$ gh pr checks 1006 --repo kewisch/ical.js
no checks reported on the 'vcard3-time-serialization' branch
```

Nothing runs on this branch, so the tables above are mine, produced locally against the published package rather than by your CI. The `main` half is read from source; I did not build the branch.

Happy to rebase, or to take the change directly and close this. No attribution needed.

PR #1029, opened 2026-09-16
Don't mangle a vCard timestamp given in the extended format; fixes #1028

### LINKED ISSUE #1028: vCard 4.0 REV timestamp in the extended format is mangled into an invalid value
### Summary

`vcardValues.timestamp` is an alias for `icalValues['date-time']` in `lib/ical/design.js`. Those conversions insert and remove the separators at fixed string offsets, which is correct for iCalendar, where the basic format is the only one. Given a vCard timestamp already in the extended format, the offsets land in the wrong places and the value is destroyed - both on the way in and on the way out, so a parse/stringify round trip silently produces an invalid vCard.

`REV` is the only property with `defaultType: "timestamp"`, so it is the one affected in practice.

### Steps to reproduce

```js
import ICAL from "ical.js";

const vcard = [
  "BEGIN:VCARD",
  "VERSION:4.0",
  "FN:Test",
  "REV:1995-10-31T22:27:10Z",
  "END:VCARD",
  "",
].join("\r\n");

const parsed = ICAL.parse(vcard);
console.log(JSON.stringify(parsed[1].find(p => p[0] === "rev")));
console.log(ICAL.stringify(parsed));
```

### Actual

```
["rev",{},"timestamp","1995--1-0-T1T:22::2"]
BEGIN:VCARD
VERSION:4.0
FN:Test
REV:1995-10-T1T22:2
END:VCARD
```

The jCal value is not a valid RFC 7095 timestamp, and the serialized `REV` is not a valid RFC 6350 timestamp either.

### Expected

The jCal value should be `1995-10-31T22:27:10Z` and `REV` should serialize to `19951031T222710Z`.

### Notes

RFC 6350's `timestamp` builds on `date-complete`, which has no separators, so strictly the input above is not conformant - the basic format is the only legal one for vCard 4.0. But the extended format is widely emitted by servers (vCard 3.0 allows it, cf. the `REV:1995-10-31T22:27:10Z` example in RFC 2426, and code carries it over to 4.0), and ical.js is lenient elsewhere in exactly this situation: `vcardValues.date` and `vcardValues.time` probe the value length and pass an already-extended value through untouched, so `BDAY`/`ANNIVERSARY` handle both formats today. `timestamp` is the outlier.

vCard 3.0 is not affected: `vcard3Properties.rev` has `defaultType: "date-time"`, and `vcard3Values['date-time']` already delegates to those conversions.

mkmelin:
The vcard4 design set aliases the timestamp value type to iCalendar's date-time, whose fromICAL/toICAL insert and remove the separators at fixed offsets because iCalendar only ever uses the basic format. RFC 6350 allows only the basic format for a timestamp as well, but the extended format is widely used in the wild, and slicing it at those offsets produced garbage: parsing REV:1995-10-31T22:27:10Z gave the jCal value 1995--1-0-T1T:22::2, stringified back as REV:1995-10-T1T22:2.

Delegate to the date-and-or-time conversions, which probe the value length instead of assuming offsets, so both formats are read and a UTC offset survives. Output stays in the basic format. decorate/undecorate are kept as they were, so getFirstValue() on REV still returns an ICAL.Time.


kewisch:
There is a `lenient` switch somewhere that should be used for more lenient parsing. The idea is that if you prefer strict validation (e.g. because you want to verify if it is correct) you can use the default, but if you want to accept usage in the wild then lenient mode is for you.

It sounds like that would be useful here? 

I think the other issue, probably outside of this PR, is that there isn't good validation in date parsing (I think back then done in the name of performance), which leads to the 1995--1-0-... problem. It should be rejected rather than producing incorrect results.

mkmelin:
Good call. The extended format is now only accepted when design.strict is false. Looks like we need to set that in Thunderbird though (for address book). 

I can open another issue for rejecting malformed date/time values

